PR #2, opened 2023-06-08
Build, load on kind and deploy makefile rules

fedepaol:
Commit on top of https://github.com/metallb/frr-k8s/pull/1

Adding everything needed to spin up a cluster and deploy in a snap, plus
pinning the executables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,22 +102,6 @@ ifndef ignore-not-found
   ignore-not-found = false
 endif
 
-.PHONY: install
-install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
-	$(KUSTOMIZE) build config/crd | kubectl apply -f -
-
-.PHONY: uninstall
-uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/crd | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
-
-.PHONY: deploy
-deploy: manifests kustomize ## Deploy controller to the K8s cluster specified in ~/.kube/config.
-	cd config/frr-k8s && $(KUSTOMIZE) edit set image controller=${IMG}
-	$(KUSTOMIZE) build config/default | kubectl apply -f -
-
-.PHONY: undeploy
-undeploy: ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/default | kubectl delete --ignore-not-found=$(ignore-not-found) -f -
 
 ##@ Build Dependencies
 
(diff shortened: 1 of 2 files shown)


hack/kind.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+set -o errexit
+
+KIND_BIN="${KIND_BIN:-kind}"
+KIND_CLUSTER_NAME="${KIND_CLUSTER_NAME:-kind}"
```
oribon, line 5: wdyt about defaulting this to frr-k8s? feels less invasive to not occupy the default cluster name :sweat_smile: (+ having a different path for this project's kubeconfig won't hurt)

Makefile
```diff
@@ -127,13 +111,35 @@ $(LOCALBIN):
...
+.PHONY: install
+install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
+	$(KUSTOMIZE) build config/crd | $(KUBECTL) apply -f -
+
+.PHONY: uninstall
+uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
+	$(KUSTOMIZE) build config/crd | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
+
+.PHONY: deploy
+deploy: manifests kustomize kind load-on-kind ## Deploy controller to the K8s cluster specified in ~/.kube/config.
```
oribon, line 135: getting a
```
... frr-k8s/bin/kubectl: No such file or directory
```
error when running this, maybe you should add a kubectl here/+ the other places that use kubectl?

Makefile
```diff
@@ -127,13 +111,35 @@ $(LOCALBIN):
...
+	cd config/frr-k8s && $(KUSTOMIZE) edit set image controller=${IMG}
+	$(KUSTOMIZE) build config/default | $(KUBECTL) apply -f -
```
oribon, line 137: nit: getting an
```
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
```

liornoy: This is because the kustomize version is 0.5+
I tried running the `kustomize edit fix` and it is not going smoothly at all.
maybe we can keep using this until it is removed completely.
from looking [here](https://github.com/kubernetes-sigs/kustomize#kubectl-integration) it looks like currently, the latest kubectl only uses kustomize v4.5.7

hack/kind.sh
fedepaol: makes sense

Makefile
```diff
@@ -127,13 +111,38 @@ $(LOCALBIN):
...
 
 ## Tool Versions
 KUSTOMIZE_VERSION ?= v5.0.0
 CONTROLLER_TOOLS_VERSION ?= v0.11.3
+KUBECTL_VERSION ?= v1.27.0
+KIND_VERSION ?= v0.19.0
+KIND_CLUSTER_NAME ?= frr-k8s
+
+.PHONY: install
+install: kubectl manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
```
oribon, line 130: nit: remove the `~/.kube/config` mentions now that we have a different kubeconfig

oribon (commented):
other than that, lgtm

Makefile
liornoy: you can change it to `bin/kubeconfig` instead of removing it completely

PR #17, opened 2023-07-03
Makefile: use go install to install kustomize

fedepaol:
Instead of using the bash script (that gets rate limited by GH) we use go install.


liornoy:
lgtm

PR #18, opened 2023-07-04
Extend the tests to cover VRFs

fedepaol:
Expanding the current tests to cover also the VRF scenario


liornoy:
lgtm. no comments :)

PR #33, opened 2023-08-22
Add metrics support

oribon:
* Fix the relevant metrics configs
* Add manifests to deploy with prometheus
* Add E2Es for metrics


hack/deploy.sh
```diff
@@ -0,0 +1,25 @@
...
+set -o errexit
+
+KUBECTL="${KUBECTL:-kubectl}"
+KUSTOMIZE="${KUSTOMIZE:-kustomize}"
+WITH_PROMETHEUS="${WITH_PROMETHEUS:-false}"
+IMG="${IMG:-quay.io/metallb/frr-k8s:latest}"
+LOGLEVEL="${LOGLEVEL:-info}"
+
+KUSTOMIZE_LAYER="default"
+if [ "$WITH_PROMETHEUS" == "true" ]; then
```
fedepaol, line 11: how about having a deploy-prometheus makefile rule instead of moving everything to this bash script?
Things will become messy once we add helm too. Also, not a fan of duplicating the env variables if not required

fedepaol:
One nit, overall looks good.
Would it make sense to add statistics related to the config status / config events as metallb does?

hack/deploy.sh
oribon: done

oribon:
> Would it make sense to add statistics related to the config status / config events as metallb does?

fedepaol:
LGTM

PR #37, opened 2023-08-28
Handle passwords

fedepaol:
Handling passwords as secret references.


internal/controller/frrconfiguration_controller.go
oribon, line 1: is there a reason we don't add a `Watches` for secrets?

config/rbac/secrets_role.yaml
oribon, line 1: does it make sense to append this to the existing role.yaml instead of a different file?

e2etests/pkg/config/from_containers.go
```diff
@@ -6,9 +6,12 @@ import (
 	"net"
 
 	frrk8sv1beta1 "github.com/metallb/frrk8s/api/v1beta1"
-	"go.universe.tf/e2etest/pkg/frr"
+	"github.com/metallb/frrk8stests/pkg/k8s"
 	frrcontainer "go.universe.tf/e2etest/pkg/frr/container"
 	"go.universe.tf/e2etest/pkg/ipfamily"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
oribon, line 13: nit: imported twice

e2etests/pkg/config/from_containers.go
```diff
@@ -40,14 +53,32 @@ func PeersForContainers(frrs []*frrcontainer.FRR, ipFam ipfamily.Family, modify
 				FRR: *f,
 			}
 
+			if f.RouterConfig.Password != "" {
```
oribon, line 56: neat

oribon:
Also, there's the `// TODO hide secrets when we have them` to fix now :)

internal/controller/api_to_config_test.go
```diff
@@ -1309,11 +1328,184 @@ func TestConversion(t *testing.T) {
...
+			expected: nil,
+			err:      errors.New("failed to process neighbor 65012@192.0.2.7 for router 65010-: secret ref not found for neighbor 65012@192.0.2.7"),
+		},
 	}
 
 	for _, test := range tests {
 		t.Run(test.name, func(t *testing.T) {
-			frr, err := apiToFRR(test.fromK8s)
+			frr, err := apiToFRR(test.fromK8s, test.secrets)
 			if test.err != nil && err == nil {
```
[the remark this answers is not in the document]
fedepaol: either that or just "expect err" as bool. Let me do that in a separate commit

config/rbac/secrets_role.yaml
fedepaol: nope, because the current role is generated by kubebuilder and gets overridden. And given the controller is mixed (mostly cluster scoped, secrets namespace scoped) kubebuilder generates a cluster role (which we don't want, as we don't want the controller to receive and read all the secrets but only the local ones)

fedepaol:
> Also, there's the `// TODO hide secrets when we have them` to fix now :)

done

e2etests/pkg/config/from_containers.go
fedepaol: done

internal/controller/api_to_config_test.go
```diff
@@ -3,19 +3,23 @@
...
+	"errors"
 	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
 	v1beta1 "github.com/metallb/frrk8s/api/v1beta1"
 	"github.com/metallb/frrk8s/internal/frr"
 	"github.com/metallb/frrk8s/internal/ipfamily"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
oribon, line 15: nit: imported twice

internal/controller/frrconfiguration_controller.go
```diff
@@ -80,7 +84,17 @@ func (r *FRRConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
 
-	config, err := apiToFRR(cfgs)
+	secrets, err := r.getSecrets(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	config, err := apiToFRR(cfgs, secrets)
+	secretNotFound := SecretNotFoundError{}
+	if errors.As(err, &secretNotFound) { // This might be a receiving order issue so we need to return the error and retry
```
oribon, line 94: is this necessary now that we listen for secret events?

oribon:
also, multiple configs for the same neighbor specifying a different password should be considered invalid?

internal/controller/frrconfiguration_controller.go
fedepaol: I'd rather be on the paranoid side. It should converge anyway

fedepaol:
> also, multiple configs for the same neighbor specifying a different password should be considered invalid?

you already added the check
https://github.com/metallb/frr-k8s/blob/7ecf5466441db43791a1337fa4745fd1dce90d6a/internal/controller/merge.go#L215

oribon (approved):
lgtm

PR #38, opened 2023-08-29
Extend metrics

oribon:
Depends on https://github.com/metallb/frr-k8s/pull/33
this aligns the ports to not clash with MetalLB and adds some metrics

### DIFF
--- a/charts/frr-k8s/templates/controller.yaml
+++ b/charts/frr-k8s/templates/controller.yaml
@@ -188,11 +188,11 @@ spec:
         args:
         - "--node-name=$(NODE_NAME)"
         - "--namespace=$(NAMESPACE)"
-        - "--metrics-bind-address=:{{ .Values.prometheus.metricsPort }}"
+        - "--metrics-bind-address={{.Values.prometheus.metricsBindAddress}}:{{ .Values.prometheus.metricsPort }}"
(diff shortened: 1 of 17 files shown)


fedepaol:
just align the helm charts too
All the rest looks fine

oribon:
> just align the helm charts too All the rest looks fine

done

charts/frr-k8s/templates/controller.yaml
```diff
@@ -188,7 +188,7 @@ spec:
         args:
         - "--node-name=$(NODE_NAME)"
         - "--namespace=$(NAMESPACE)"
-        - "--metrics-bind-address=127.0.0.1:{{ .Values.prometheus.metricsPort }}"
+        - "--metrics-bind-address=:{{ .Values.prometheus.metricsPort }}"
```
fedepaol, line 191: I think binding to 127.0.0.1 is better than listening on any

oribon: I changed it here so it is aligned with the regular manifests and the code's default (when we generated the controller), do you think we should change it everywhere?

fedepaol: yeah I think this is better as the pod is hostnetworked adn we won't expose the port outside.

fedepaol:
LGTM

PR #40, opened 2023-09-01
Allow injecting raw configurations

fedepaol:
To leverage the frr instance and explore new use cases, we allow injecting raw configs that are appended after the rendered one.
Given that multiple configurations might be applied to the same node, a priority field drives which config should go first.


api/v1beta1/frrconfiguration_types.go
```diff
@@ -26,19 +26,29 @@ type FRRConfigurationSpec struct {
...
 	// whose labels match the specified selectors.
 	// When it is not specified all nodes will attempt to apply this config.
 	// +optional
 	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty"`
+}
+
+type RawConfig struct {
+	// Sets the order with this configuration is appended to the
+	// bottom of the rendered configuration
+	Priority int `json:"priority,omitempty"`
```
oribon, line 42: does this have a value for the user (other than us using this for "deterministic order"?)

internal/controller/api_to_config.go
```diff
@@ -346,3 +352,15 @@ func sortMapPtr[T any](toSort map[string]*T) []*T {
 	}
 	return res
 }
+
+func joinRawConfigs(raw []v1beta1.RawConfig) string {
+	sort.Slice(raw, func(i, j int) bool {
+		return raw[i].Priority > raw[j].Priority
```
oribon, line 358: this is nondeterministic if `raw` contains multiple with the same prio in a different order everytime (maybe it makes sense to sort the fromK8s cfgs first by name)?

internal/controller/api_to_config_test.go
```diff
@@ -1309,6 +1309,81 @@ func TestConversion(t *testing.T) {
...
+						IPV4Prefixes: []string{},
+						IPV6Prefixes: []string{},
+					},
+				},
+				ExtraConfig: "foo\n",
+			},
+			err: nil,
+		},
+		{
+			name: "Single Router and double injection",
```
oribon, line 1347: is there any value in testing a multi line raw config?

e2etests/tests/raw_config.go
```diff
@@ -0,0 +1,183 @@
...
+
+			nodes, err := k8s.Nodes(cs)
+			framework.ExpectNoError(err)
+
+			for _, c := range frrs {
+				ValidateFRRPeeredWithNodes(nodes, c, ipfamily.IPv4)
+			}
+			config.Spec.Raw.Config = nil
+			err = updater.Update(config)
+			framework.ExpectNoError(err)
```
oribon, line 129: what does this do?

api/v1beta1/frrconfiguration_types.go
fedepaol: well the order (and having the control of what config lands) is not only for us.
They might want to define the rotuer in one config and the rules in another, or what not.
It makes sense to control the order.

internal/controller/api_to_config_test.go
fedepaol: I can add

e2etests/tests/raw_config.go
fedepaol: Leftover, let me remove it

oribon (commented):
small nit, other than that lgtm

api/v1beta1/frrconfiguration_types.go
```diff
@@ -26,19 +26,29 @@ type FRRConfigurationSpec struct {
...
 	// When specified, the configuration will be considered only on nodes
 	// whose labels match the specified selectors.
 	// When it is not specified all nodes will attempt to apply this config.
 	// +optional
 	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty"`
+}
+
+type RawConfig struct {
+	// Sets the order with this configuration is appended to the
+	// bottom of the rendered configuration
```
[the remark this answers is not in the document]
fedepaol: *before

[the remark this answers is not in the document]
fedepaol: done

oribon (approved):
lgtm

PR #53, opened 2023-10-05
Don't treat the non existing secret as a special error

fedepaol:
We do listen for secrets, so if it is added it the reconciliation will converge. By returning the error we add a lot of unnecessary loops happening before the secret is added.


oribon (approved):
lgtm

PR #54, opened 2023-10-06
Translate the keepalive / holdtime from api to frr

fedepaol:
Those fields were missing. Also, we do validate their values as we did in metallb.

Note: this requires https://github.com/metallb/metallb/pull/2104 to make the tests compile

### DIFF
--- a/e2etests/tests/session.go
+++ b/e2etests/tests/session.go
@@ -1,0 +1,118 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package tests
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+
+	frrk8sv1beta1 "github.com/metallb/frrk8s/api/v1beta1"
+	"github.com/metallb/frrk8stests/pkg/config"
+	"github.com/metallb/frrk8stests/pkg/dump"
+	"github.com/metallb/frrk8stests/pkg/infra"
+	"github.com/metallb/frrk8stests/pkg/k8s"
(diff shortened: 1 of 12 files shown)


internal/controller/api_to_config_test.go
```diff
@@ -295,7 +305,8 @@ func TestConversion(t *testing.T) {
 								Name:     "65031@192.0.2.16",
 								ASN:      65031,
 								Addr:     "192.0.2.16",
-								Port:     179,
+
```
oribon: nit: unrelated

e2etests/tests/session.go
```diff
@@ -0,0 +1,118 @@
...
+								ASN: infra.FRRK8sASN,
+								Neighbors: []frrk8sv1beta1.Neighbor{
+									{
+										ASN:     1234,
+										Address: "192.168.1.1",
+										HoldTime: metav1.Duration{
+											Duration: 90 * time.Second,
+										},
+										KeepaliveTime: metav1.Duration{
+											Duration: 30 * time.Second,
```
oribon, line 82: is there value in using something other than the defaults?

oribon (approved):
small nits, lgtm

e2etests/tests/session.go
fedepaol: there is, as by creating with the api doesn't use the CRD default values. However, slightly changing the values so it offers better coverage

internal/controller/api_to_config_test.go
fedepaol: fixed

PR #52, opened 2023-10-04
Expose the current FRR status

fedepaol:
We add a new CRD which will expose the following information:

- the current running configuration of an frr instance
- the desired configuration that the daemon wants to apply
- the result of the last desired configuration apply, which might be
  success or the failure message

PR #63, opened 2023-10-18
API: add a LastConversionResult field to the node state

fedepaol:
This might be useful to understand why the conversion failed without having to look at the logs. The FRR status is useful, but sometimes the configuration is blocked earlier (for example when the password is missing from secrets).

Based off https://github.com/metallb/frr-k8s/pull/52

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -163,24 +163,27 @@ func main() {
 		}
 		frrInstance := frr.NewFRR(ctx, reloadStatus, logger, logging.Level(logLevel))
 
-		if err = (&controller.FRRConfigurationReconciler{
-			Client:     mgr.GetClient(),
-			Scheme:     mgr.GetScheme(),
-			FRRHandler: frrInstance,
-			Logger:     logger,
-			NodeName:   nodeName,
-		}).SetupWithManager(mgr); err != nil {
+		configReconciler := &controller.FRRConfigurationReconciler{
+			Client:       mgr.GetClient(),
+			Scheme:       mgr.GetScheme(),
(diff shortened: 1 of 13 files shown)


cmd/main.go
```diff
@@ -163,24 +163,27 @@ func main() {
...
-			Logger:    logger,
-			NodeName:  nodeName,
-			Update:    reloadStatusChan,
+			Client:           mgr.GetClient(),
+			Scheme:           mgr.GetScheme(),
+			FRRStatus:        frrInstance,
+			Logger:           logger,
+			NodeName:         nodeName,
+			Update:           reloadStatusChan,
+			ConversionResult: configReconciler,
```
oribon, line 186: nit/general thought: the controllers are all in the same pkg, is there a reason not too have these as vars on the pkg scope? I feel it is more clunky this way (main caring that a particular controller is a "conversionFetcher" and should be passed to the other) rather than just letting the relation happen on the pkg. wdyt?

fedepaol: If I got the comment right...

Using global variables is a shortcut that removes any form of encapsulation. It's like putting pipes between the otherwise self contained objects. 

The `ConversionResult` field is an external dependency that this controller has with something that lives outside of it, which _incidentally_ is another controller living in the same package. As an external dependency, it must come from outside and injected to the object.

This way is more modular and sets a clear boundary between the various objects.  The extra cost, as you said, is the wiring / injection of the various dependencies but I think this way is cleaner.

oribon (approved):
lgtm

PR #69, opened 2023-11-01
Add a mechanism to reset the FRR config if the node breaks

oribon:
We add a new controller that runs a healthcheck periodically and every time we change the frr config, if it fails the frr config is reset and the node status is updated.


internal/frr/frr.go
```diff
@@ -34,38 +33,24 @@ type Status struct {
 
 type FRR struct {
 	reloadConfig    chan reloadEvent
-	logLevel        string
```
fedepaol, line 37: this was supposed to exist as a handle for changing teh configuration on the fly, but since we don't have such option today I guess it's better to remove it and reintroduce it when we have the option

internal/controller/frrconfiguration_controller.go
```diff
@@ -156,15 +156,15 @@ func (r *FRRConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	return ctrl.Result{}, nil
 }
 
-func (r *FRRConfigurationReconciler) applyEmptyConfig(req ctrl.Request) error {
+func (r *FRRConfigurationReconciler) ApplyEmptyConfig() error {
```
fedepaol, line 169: Let's add the req to the external error if it fails

internal/controller/healthchecker.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func NewHealthCheckEvent() event.GenericEvent {
+	evt := hcEvent{}
+	evt.Namespace = "healthcheck"
+	evt.Name = "check"
+	return event.GenericEvent{Object: &evt}
+}
+
+func NewHealthChecker(cl client.Reader, update chan event.GenericEvent, l log.Logger,
```
fedepaol, line 69: Let's call this HealthCheckController because it's a controller (same for the file name)

internal/controller/healthchecker.go
```diff
@@ -0,0 +1,161 @@
...
+		r.healthResMutex.Lock()
+		r.healthResult = healthResult
+		r.healthResMutex.Unlock()
+		if healthResult != lastHealthResult {
+			r.ReloadStatus()
+		}
+	}()
+
+	var err error
+	for i := 0; i < r.attempts; i++ {
```
fedepaol, line 103: Does it make sense to have the attempts as part of the check itself?

internal/controller/healthchecker.go
```diff
@@ -0,0 +1,161 @@
...
+		err = r.check()
+		if err == nil {
+			break
+		}
+	}
+	if err == nil {
+		return ctrl.Result{}, nil
+	}
+	level.Error(r.Logger).Log("controller", "HealthChecker", "failed, error: ", err)
```
fedepaol, line 112: Let's log the fact that we are not able to reach the apiserver and that we will reset the FRR configuration

internal/controller/healthchecker.go
```diff
@@ -0,0 +1,161 @@
...
+
+	err = r.ApplyEmptyConfig.ApplyEmptyConfig()
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	healthResult = time.Now().Format(time.DateTime)
```
fedepaol, line 118: This is quite hard to follow. We are setting here so the defer function finds the differences. Can we just handle it without the defer? After all we need to re-assign only if we are in the last leg of the function.
That would also remove the need for checking if it changed I think

internal/controller/healthchecker.go
```diff
@@ -0,0 +1,161 @@
...
+	configStale.Set(1) // the config controller would set it 0 the next time a valid config is applied
+	configReset.Inc()
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *HealthChecker) SetupWithManager(mgr ctrl.Manager) error {
+	if r.interval > 0 {
+		go r.feedUpdate()
```
fedepaol, line 129: we should pass the outer context here so the goroutine can stop

internal/controller/stats.go
```diff
@@ -38,9 +38,16 @@ var (
 		Name:      "config_stale_bool",
 		Help:      "1 if running on a stale configuration, because the latest config failed to load.",
 	})
+
+	configReset = prometheus.NewCounter(prometheus.CounterOpts{
+		Namespace: Namespace,
+		Subsystem: Subsystem,
+		Name:      "config_reset_total",
+		Help:      "Number of times the healthcheck controller reset the config.",
```
fedepaol, line 46: Does it make sense to have this as a gauge (0 - 1) to say if the configuration was reset?

api/v1beta1/frr_node_state_types.go
```diff
@@ -30,6 +30,7 @@ type FRRNodeStateStatus struct {
 	RunningConfig        string `json:"runningConfig,omitempty"`
 	LastConversionResult string `json:"lastConversionResult,omitempty"`
 	LastReloadResult     string `json:"lastReloadResult,omitempty"`
+	LastHealthcheckReset string `json:"lastHealthcheckReset,omitempty"`
```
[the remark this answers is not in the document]
oribon: this is what I initially did, but the problem with a boolean (or a "success" string) is that the controller needs to be aware of what config FRR is running when reconciling, because otherwise the field will be flipped next time the health check runs on an empty config. I thought this is fine to not over-complicate what's going on here and provides enough info ("intersecting" with the other status fields), but it's probably possible to expose the FRRHandler's last applied config and consume it here (by comparing in which "state" the current reconcile is). wdyt?

internal/controller/frrconfiguration_controller.go
oribon: done

internal/controller/healthchecker.go
oribon: done

internal/controller/healthchecker.go
oribon: I think keeping this here is cleaner (the check being focused on doing its thing), also might be better if we decide later to add more checks. anyways I don't have a strong preference at this moment, wdyt?

internal/controller/healthchecker.go
oribon: done

internal/controller/healthchecker.go
oribon: yep (was a leftover)

internal/controller/healthchecker.go
oribon: it uses the context that is passed to the manager (`r.ctx.Done()`) to stop, do you think I should change it so it is  part of the Setup instead of a field on the reconciler?

internal/controller/stats.go
oribon: this depends on the other discussion

e2etests/tests/status.go
```diff
@@ -255,6 +260,114 @@ var _ = ginkgo.Describe("Exposing FRR status", func() {
...
+			}
+
+			controlPlaneNodes := corev1.NodeList{}
+			cpLabels := map[string]string{"node-role.kubernetes.io/control-plane": ""}
+			err = cl.List(context.Background(), &controlPlaneNodes, &client.ListOptions{LabelSelector: labels.SelectorFromSet(cpLabels)})
+			framework.ExpectNoError(err)
+			Expect(controlPlaneNodes.Items).ToNot(BeEmpty())
+
+			raw := ""
+			for _, n := range controlPlaneNodes.Items {
```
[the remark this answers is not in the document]
oribon: done

e2etests/tests/status.go
```diff
@@ -255,6 +260,114 @@ var _ = ginkgo.Describe("Exposing FRR status", func() {
...
 			}
 		})
+
+		ginkgo.It("Works with a config that breaks worker nodes api-server connectivity", func() {
+			states := frrk8sv1beta1.FRRNodeStateList{}
+			err := cl.List(context.Background(), &states)
+			framework.ExpectNoError(err)
+
+			// The nodes that we want to use for this test are non-control planes
+			// that do not host a webhook endpoint to minimize disruption to the cluster.
```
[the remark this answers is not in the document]
oribon: done

e2etests/tests/status.go
```diff
@@ -255,6 +260,114 @@ var _ = ginkgo.Describe("Exposing FRR status", func() {
...
+			}
+
+			whPods := corev1.PodList{}
+			whLabels := map[string]string{"app.kubernetes.io/component": "frr-k8s-webhook-server"}
+			err = cl.List(context.Background(), &whPods, &client.ListOptions{
+				Namespace:     k8s.FRRK8sNamespace,
+				LabelSelector: labels.SelectorFromSet(whLabels)})
+			framework.ExpectNoError(err)
+
+			for _, p := range whPods.Items {
```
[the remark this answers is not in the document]
oribon: done

api/v1beta1/frr_node_state_types.go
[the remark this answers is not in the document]
fedepaol: I understand the complication, but we need ot have a way to understand if the current status is "reset because of healthcheck".
Another possible approach is to have the configuration controller to reset the state whenever it applies a new configuration (because in the next status update we would want to see the flag as valid until the next healthcheck).

However, I am not sure I like the number of arrows that we are drawing between all the controllers. 
Basically we have the "status" of the node that is composed of:
- what we get from frr (which kind of belongs to the status controller)
- the conversion failed information, which belongs to the configuration controller
- the healthcheck failure (which belongs the the healthcheck controller and it MAY be contributed by the configuration controller)

I have a feelng that things could be much smoother if we merge the health check and the configuration controller, as they both mess up with the configuration.

internal/controller/healthcheck_controller.go
```diff
@@ -0,0 +1,153 @@
...
+		if err == nil {
+			break
+		}
+	}
+	if err == nil {
+		return ctrl.Result{}, nil
+	}
+	level.Error(r.Logger).Log("controller", "Healthcheck", "failed to reach apiserver", "resetting FRR config", "err", err)
+
+	err = r.ApplyEmptyConfig.ApplyEmptyConfig(req)
```
fedepaol, line 102: I just realized, this might cause races between a configuration coming from outside and a failed healthcheck.
There is nothing protecting it. As I wrote in the other comment, I feel merging this two controllers might remove a lot of headaches

PR #71, opened 2023-11-08
Set optional FRR fields as pointers

oribon:
We set the FRR parameters as pointers as we want to defer default values to FRR (and have a cleaner config).


fedepaol:
LGTM

PR #72, opened 2023-11-08
Enforce router prefixes on neighbors

oribon:
As stated in the API we make sure the AllowedOut prefixes match the prefixes defined in the router.


fedepaol:
LGTM

PR #77, opened 2023-11-27
Charts: disable service monitors by default

fedepaol:
We don't know if the dest cluster has the prometheus operator deployed, so it's not safe to assume so by default.

Also, removing comments related to metallb alarms.


oribon (approved):
lgtm

PR #82, opened 2023-12-05
dump: hide cleartext passwords

oribon:
We forgot to hide cleartext passwords when we added them, we also remove the `kubectl.kubernetes.io/last-applied-configuration` annotation from the dumped config as it contains a password + redundant data.


fedepaol:
lgtm

PR #86, opened 2023-12-12
controller: fix merging neighbor's timers

oribon:
We accidentally compared pointers instead of their values.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,10 @@
 
 ## Next release
 
+### Bug fixes
+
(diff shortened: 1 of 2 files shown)


RELEASE_NOTES.md
```diff
@@ -2,6 +2,10 @@
 
 ## Next release
 
+### Bug fixes
+
+- Fix comparing neighbor's timers pointers instead of their values.
```
fedepaol, line 7: can you put the PR # and a more user friendly description of the bug (i.e. merging always fails if holdtime / keepalive time is set)?

oribon: done

fedepaol (approved):

PR #88, opened 2023-12-14
Add a parameter to define the cidrs to always block

fedepaol:
In order to protect the cluster, we set a parameter where the user can define a list of cidrs to always block.

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,7 @@ spec:
               - prefix: 192.168.1.0/24
               - prefix: 192.169.2.0/24
                 ge: 25
-                le: 28   
+                le: 28
 ```
 
 le / ge specify prefix length. The prefix will be applied if the prefix length is less than or equal to the le prefix length
@@ -195,7 +195,7 @@ spec:
           port: 179
           toReceive:
             allowed:
-              mode: all 
+              mode: all
 ```
 
 ### Adding a raw configuration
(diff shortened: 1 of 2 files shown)


internal/frr/templates/filters.tmpl
```diff
@@ -75,9 +75,22 @@ ip prefix-list {{$plistName}} seq {{counter $plistName}} deny any
 {{- end }}
 {{ if not .neighbor.Incoming.PrefixesV6 }}
 ipv6 prefix-list {{$plistName}} seq {{counter $plistName}} deny any
+{{- end }}
+
+{{- if .neighbor.AlwaysBlock}}
+
+{{ $plistName:=deniedIncomingList $.neighbor}}
+{{- range $a := .neighbor.AlwaysBlock }}
+{{frrIPFamily $a.IPFamily}} prefix-list {{$plistName}} seq {{counter $plistName}} permit {{$a.Prefix}}{{$a.Matcher}}
```
oribon, line 84: nit/question: why is this permit and not deny?

internal/controller/api_to_config.go
```diff
@@ -523,6 +525,20 @@ func joinRawConfigs(raw []namedRawConfig) string {
 	return res.String()
 }
 
+func alwaysBlockToFRR(cidrs []net.IPNet) []frr.IncomingFilter {
+	res := make([]frr.IncomingFilter, 0, len(cidrs))
+	for _, c := range cidrs {
+		family := ipfamily.ForCIDR(&c) // #nosec G601
```
oribon, line 531: why use nosec?

oribon (commented):
few nits, other than that lgtm

internal/frr/templates/filters.tmpl
fedepaol: permit, in the context of a prefix list, means the prefix is added to the prefix list. If we were adding deny, then the prefix would not be added and thus ignored by the rule

internal/controller/api_to_config.go
fedepaol: the warning is about the pointer being referenced in a loop. The risk, I guess, is to carry around the pointed object which changes while looping. In this case, there's no risk as ForCIDR is a pure function that does not depend on &c but just returns its family. We could've added a local copy to make gosec happy but imo that would've been a waste of resources.

PR #91, opened 2023-12-19
webhook: validate resources from all namespaces

oribon:
We accidently validated only against resources from the namespace the daemon is deployed on. This fixes it and modifies the relevant e2e.


fedepaol:
LGTM

PR #93, opened 2023-12-20
helm: add an option to disable the webhook's cert rotation

oribon:
We add an option to disable the webhook's cert rotator, this enables another entity to generate and manage the relevant secret.


fedepaol:
LGTM

PR #97, opened 2023-12-25
helm: align webhooks probes with kustomize

oribon:
The liveness/readiness probes of the webhook pod pointed to the wrong endpoints,
essentially checking the endpoint of the daemon pod (via 127.0.0.1).
This fixes this by aligning with what we do in kustomize, probing the
pod's correct livez/healthz endpoints.


fedepaol:
LGTM :santa:

PR #99, opened 2023-12-31
CI: Add a MetalLB E2E lane

oribon:
We add a MetalLB lane to make sure we aren't introducing changes that break the integration between the two projects.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -187,3 +187,92 @@ jobs:
         uses: ./.github/workflows/composite/collectlogs
         with:
           artifact-name: kind-logs-${{ matrix.ip-family }}-${{ matrix.deployment }}
+
+  metallb_e2e:
+    runs-on: ubuntu-22.04
+    needs:
+      - unit-tests
+      - build-test-images
+      - commitlint
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install linux-modules-extra-$(uname -r) python3-pip arping ndisc6
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -187,3 +187,92 @@ jobs:
...
+          mkdir -p /tmp/kind_logs/
+
+      - name: Checkout MetalLB
+        uses: actions/checkout@v4
+        with:
+          repository: metallb/metallb
+          path: metallb
+          ref: "main"
+
+      - name: Deploy MetalLB
```
fedepaol, line 238: Why not just deploying metallb in frr mode overriding the frrk8s image? What am I missing?

oribon: I think it would be messy when prs change the helm chart itself (calling the new changes from metallb's chart)

fedepaol: I _think_ we could just copy the whole chart folder under metallb's one (which is what we end up doing when we declare frrk8s as a metallb dependency anyway). That'd remove all the logic we are adding here.
Does it make sense? I think it's worth an attempt

oribon: sure I can try

oribon: it looks like I made it work, but not sure I like the outcome (the one later where we will use `inv dev-env`) 😅 seems a bit hacky (unless that's not what you meant), and will also require updating dev-env to accept extra helm args + loading more images into kind. wdyt should we continue with this or revert to what I originally did?

fedepaol: yep let's revert and see if this won't bite us in the future :P

oribon: done

PR #100, opened 2024-01-11
Add option to enable the rotator RestartOnSecretRefresh

oribon:
When the webhook pod is redeployed multiple times in succession sometimes the requests for the webhook hang on `tls: failed to verify certificate` and it is not solved until the pod is restarted manually. While the reason the flag solves the issue is not totally clear, introducing the flag for the operator to consume is reasonable.


fedepaol:
LGTM

PR #107, opened 2024-01-19
Add a demo creation script

fedepaol:
We setup an external frr container and provide multiple configurations to showcase frr-k8s features.


hack/demo/demo.sh
oribon, line 1: iiuc this relies on first creating a kind cluster with everything? could/should this be smoother by first deploying from the makefile target?

hack/demo/configs/templates/advertise.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: frrk8s.metallb.io/v1beta1
+kind: FRRConfiguration
+metadata:
+  name: advsome
```
oribon, line 4: nit: rename these to something clearer - advertise-filtered, advertise-all, nodeselectors, ..

hack/demo/demo.sh
```diff
@@ -0,0 +1,30 @@
...
+
+for i in configs/*.yaml; do
+	rm $i
+done
+
+cp configs/templates/*.yaml configs/
+
+for i in configs/*.yaml; do
+    sed -i "s/NEIGHBOR_IP/$FRR_IP/g" $i
+done
```
oribon, line 30: can we also echo here that the setup is complete, yamls are in `hack/demo/..` ?

hack/demo/demo.sh
fedepaol: done

hack/demo/configs/templates/advertise.yaml
fedepaol: done

hack/demo/demo.sh
fedepaol: done

hack/demo/frr/frr.conf
oribon, line 1: nit: I think this can be .gitignored ?

oribon (commented):
other than that lgtm

PR #110, opened 2024-01-20
API: remove desired state

fedepaol:
The desired state field is too verbose and does not provide much difference from the current one (errors aside).
Removing it will make the API more slick in my opinion, but open to discussion.

cc @oribon


oribon (approved):
lgtm

PR #111, opened 2024-01-20
CI: don't check generated files for dependabot prs

fedepaol:
Assuming dependabot knows how to update a dependency, we don't want to check changes in go.mod (but also, all the generated files checks we do) on dependabot PRs.

So, we split the lint lane in a "lint" job which performs only linting, and a more generic "check generated" that does not run when dependabot files a PR.


oribon (approved):
lgtm

PR #112, opened 2024-01-20
Makefile: ignore kubectl errors when checking the version

fedepaol:
If there is no connectivity to the cluster, kubectl version fails and so the version check does not even get to check the version.

Here we redirect stderror so we can check the version and avoid downloading every time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -200,7 +200,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: kubectl
 kubectl: $(KUBECTL) ## Download kubectl locally if necessary. If wrong version is installed, it will be overwritten.
 $(KUBECTL): $(LOCALBIN)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -200,7 +200,7 @@ $(CONTROLLER_GEN): $(LOCALBIN)
 .PHONY: kubectl
 kubectl: $(KUBECTL) ## Download kubectl locally if necessary. If wrong version is installed, it will be overwritten.
 $(KUBECTL): $(LOCALBIN)
-	test -s $(LOCALBIN)/kubectl && $(LOCALBIN)/kubectl version | grep -q $(KUBECTL_VERSION) || \
+	test -s $(LOCALBIN)/kubectl && $(LOCALBIN)/kubectl version 2> /dev/null | grep -q $(KUBECTL_VERSION) || \
```
oribon, line 203: can we use `kubectl version --client` instead?
```
    --client=false:
	If true, shows client version only (no server required).
```

PR #120, opened 2024-02-22
Merge: don't convert the default values from seconds

fedepaol:
The target is already expressed in seconds, so we must
consider the default directly expressed in seconds.

### DIFF
--- a/internal/controller/merge.go
+++ b/internal/controller/merge.go
@@ -4,10 +4,15 @@ package controller
 
 import (
 	"fmt"
-	"time"
 
 	"github.com/metallb/frr-k8s/internal/frr"
 	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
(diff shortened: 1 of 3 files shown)


oribon (commented):
lgtm other than that

oribon (approved):
lgtm

internal/controller/merge.go
```diff
@@ -207,7 +226,7 @@ func neighborsAreCompatible(n1, n2 *frr.NeighborConfig) error {
 		return fmt.Errorf("multiple asns specified for %s", neighborKey)
 	}
 
-	if !ptrsEqual(n1.Port, n2.Port, 179) {
+	if !ptrsEqual(n1.Port, n2.Port, defaultBGPPort) {
```
oribon, line 229: nit: ptrsEqual might be redundant now

PR #135, opened 2024-04-05
Pin go version, update ci workflow and bump controller_tool version

karampok:
Currently we use the go version which is 1.21 in the ubuntu image from the runner. That probably will change soon and ci "probably" will start failing.


.github/workflows/ci.yaml
```diff
@@ -40,7 +40,10 @@ jobs:
         uses: actions/checkout@v4
         with:
           fetch-depth: 0
-
+      - uses: actions/setup-go@v5
```
fedepaol, line 43: we should do the same for the other jobs I think (e2e / metallb e2e)

karampok: e2e they use  only `make e2etests` and then  `$(GINKGO) -v $(GINKGO_ARGS) --timeout=3h ./e2etests -- $(TEST_ARGS)`  so  we should control the ginkgo version probably by doing `go install -mod=vendor github.com/onsi/ginkgo/v2/ginkgo` (need to check if we do vendor). Not sure the version of go is important or not

Probably same for metallb2 e2e. 

But we can add it

fedepaol: go install compiles gingko, so yes, in order to have less drifts as possible I think it make sense to pin there too.
The ginkgo version is pinned already.

fedepaol:
Thanks! Just one thing: in this case I'd split at least the part where we bump controller-gen and all the generated bits, and reword the commit title (and the PR title) as we are doing much more than advertised in the title / subject

karampok:
Do i need to "update branch" from github(or from my local fork), so this PR to be ready to be merged?

fedepaol:
> Do i need to "update branch" from github(or from my local fork), so this PR to be ready to be merged?

nope, it's not required as we have merge queues activated. And even if we didn't have, the maintiners have that option so it's better to leave the choise to rebase to them (because there might be multiple prs to be merged, and the rebasing order might differ)

fedepaol:
LGTM, sending to merge queue

fedepaol:
> Thanks! do you know why there are some borken github actions workflows around the generated files?

This is the publish workflow, so it could be that we have some difference there (such as the go version used?)

Issue #132, opened 2024-03-31
`SrcAddr` setting is not wired into the API


MatthiasValvekens:
Hi, there seems to be an internal `SrcAddr` setting to populate `update-source` for neighbors that is not exposed in the K8S API. Is that intentional?

I think my use case would benefit from it. I'm trying to expose some routes to IPv6 load balancers from multiple nodes in my cluster, and to do that I'm peering with an IPv6 router. Now, the interface that is connected to that router has many IPv6 addresses (on all nodes involved), so I need to make sure that the one selected by the kernel matches the one that is used to establish the BGP session from the peer end. This leads me to the following problem:

- In MetalLB in "regular" FRR mode, I can't seem to define two `BGPPeer` resources with the same neighbor address (even with mutually exclusive node selectors) because the API forbids that, so I can only align the source address for at most one node.
- With `frrk8s.enabled=true`, I thought I could get away with a single `BGPPeer` for all nodes and override the source address on a per-node basis by introducing extra `FRRConfiguration` resources scoped to specific nodes. But since this setting is apparently not exposed in the API, the `sourceAddress` feature doesn't work at all in this setup.

Is that something that simply hasn't been implemented yet, or is there something fundamental I'm not understanding here?

Since I control all the devices involved, I can probably work around this problem by giving extra IPs to the interface on the peer router's end, allowing me to create one BGPPeer per node in my cluster, but that feels like a hack. Thoughts?

Thank you!

oribon:
> Is that intentional?

probably not, we pretty much copied the internal structs and did the adjustments, there's a chance we missed implementing this😅 I don't remember we had some discussion about not implementing this.
even if we do implement this we'll need to think if the merging is right - that is giving the priority to a specified srcAddr (by an additional config) over an empty one (by MetalLB).

ccing @fedepaol maybe he remembers something about this (he should be back in a few days )

Thanks a lot for raising this!

MatthiasValvekens:
Thanks!

> that is giving the priority to a specified srcAddr (by an additional config) over an empty one (by MetalLB).

Just as a point of clarification: MetalLB also allows specifying `sourceAddress` at the `BGPPeer` level (and that works in FRR mode), but this setting is also not applied with frr-k8s AFAICT (which would make sense, since MetalLB generates input for the k8s API).

oribon:
yep we're on the same page 😄 nothing is implemented regarding frr-k8s and source addresses

MatthiasValvekens:
[...]

PR #137, opened 2024-04-10
Expose src addr in the k8s API

karampok:
/kind feature

```release-note
Expose the IPv4 or IPv6 source address to use for the BGP session through the k8s API
```

Fixes https://github.com/metallb/frr-k8s/issues/132

### DIFF
--- a/API-DOCS.md
+++ b/API-DOCS.md
@@ -226,6 +226,7 @@ _Appears in:_
 | Field | Description |
 | --- | --- |
 | `asn` _integer_ | ASN is the AS number to use for the local end of the session. |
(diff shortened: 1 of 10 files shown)


fedepaol:
I see this (silently) contains changes from https://github.com/metallb/frr-k8s/pull/135.
When we do something like this (and only if required) we just bring the same set of commits and declare the PR to contain the other changes as well.

fedepaol:
Can you also add a test on the frr pkg side to see that the configuration is rendered correctly?

karampok:
The other PR is not really need, I was expected that would have been merged. I will remove the silent changes and will add the test in the frr pkg side (might ask for help if cannot figure out the right place).

fedepaol:
Thanks!

PR #162, opened 2024-06-26
API: add graceful restart support

karampok:
/kind feature

```release-note
Add graceful restart support
```


karampok:
@fedepaol please have a look, I think is complete. Thank you

e2etests/tests/session.go
```diff
@@ -123,7 +123,7 @@ var _ = ginkgo.Describe("Advertisement", func() {
 			}
 		})
 
-		ginkgo.DescribeTable("Establishes sessions with cleartext password", func(family ipfamily.Family) {
+		ginkgo.DescribeTable("Establishes sessions with cleartext password and graceful restart", func(family ipfamily.Family) {
```
fedepaol, line 126: can we have a separate test instead?
Also, is there benefit in testing e2e (also, looking at the other side when frr-k8s receive routes) as we did in metallb?

api/v1beta1/frrconfiguration_types.go
```diff
@@ -143,6 +143,12 @@ type Neighbor struct {
 	// +optional
 	BFDProfile string `json:"bfdProfile,omitempty"`
 
+	// EnableGracefulRestart allows BGP peer to continue to forward data packets along
```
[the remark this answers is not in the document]
karampok: yeah, I want to discuss that with you.
I tried but when I deploy `make deploy-controller` I get this error
```
The CustomResourceDefinition "frrconfigurations.frrk8s.metallb.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[bgp].properties[routers].items.properties[neighbors].items.properties[enableGracefulRestart].x-kubernetes-validations[0].rule: Invalid value: "self == oldSelf": oldSelf cannot be used on the uncorrelatable portion of the schema within spec.validation.openAPIV3Schema.properties[spec].properties[bgp].properties[routers]
```
which looks we can not use this in `uncorrelatable portion of the schema`. Maybe I am doing something wrong not sure

https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/validation/validation.go#L852

Not sure what is the way forward (if any)
1. custom logic to execute a reset on the bgp neighbor if this flag is added
2. custom logic to just show degraded status/warning
3. any other good idea?

e2etests/tests/session.go
karampok: > can we have a separate test instead?
I will do that sure

> is there benefit in testing e2e (also, looking at the other side when frr-k8s receive routes) as we did in metallb?
I would like first to connect metallb/frr-k8s and make the the we did there to run also in the FRR-k8s mode.
Mixed feeling if there is benefit but no objection to add. Do you have any test I can replicate or should I build from scratch?

api/v1beta1/frrconfiguration_types.go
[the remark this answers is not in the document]
fedepaol: > yeah, I want to discuss that with you. I tried but when I deploy `make deploy-controller` I get this error
> 
> ```
> The CustomResourceDefinition "frrconfigurations.frrk8s.metallb.io" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[bgp].properties[routers].items.properties[neighbors].items.properties[enableGracefulRestart].x-kubernetes-validations[0].rule: Invalid value: "self == oldSelf": oldSelf cannot be used on the uncorrelatable portion of the schema within spec.validation.openAPIV3Schema.properties[spec].properties[bgp].properties[routers]
> ```
> 
> which looks we can not use this in `uncorrelatable portion of the schema`. Maybe I am doing something wrong not sure
> 
> https://github.com/kubernetes/apiextensions-apiserver/blob/master/pkg/apis/apiextensions/validation/validation.go#L852
> 
> Not sure what is the way forward (if any)
> 
>     1. custom logic to execute a reset on the bgp neighbor if this flag is added
> 

This is bad, as the user would declare a desired state that not only we can't fulfill, but we also change without him knowing.

>     2. custom logic to just show degraded status/warning
> 

We could have the validation done both in the webhook / config loading, but it's not ideal

>     3. any other good idea?

Looking at the source code!

The error comes from here

https://github.com/kubernetes/apiextensions-apiserver/blob/5d129eee2b85758751c8a23f64ff2d647dd48916/pkg/apis/apiextensions/validation/validation.go#L1261

It fails if 

```
if uncorrelatablePath := ssv.forbidOldSelfValidations(); uncorrelatablePath != nil {
```

which comes from here https://github.com/kubernetes/apiextensions-apiserver/blob/5d129eee2b85758751c8a23f64ff2d647dd48916/pkg/apis/apiextensions/validation/validation.go#L1397:

which is being set on 

https://github.com/kubernetes/apiextensions-apiserver/blob/5d129eee2b85758751c8a23f64ff2d647dd48916/pkg/apis/apiextensions/validation/validation.go#L1075


```
		// we can only correlate old/new items for "map" and "set" lists, and correlation of
		// "set" elements by identity is not supported for cel (x-kubernetes-validations)
		// rules. an unset list type defaults to "atomic".
```

It basically means, to make assumptions on the old value we must be able to understand who is what.

I made it work by:

- declaring the neighbor list as map, with the address as key
- declaring the router list as a map, with vrf as key


Diff as follows:

```
diff --git a/api/v1beta1/frrconfiguration_types.go b/api/v1beta1/frrconfiguration_types.go
index 430e88e..8ec9c32 100644
--- a/api/v1beta1/frrconfiguration_types.go
+++ b/api/v1beta1/frrconfiguration_types.go
@@ -56,6 +56,8 @@ type RawConfig struct {
 type BGPConfig struct {
        // Routers is the list of routers we want FRR to configure (one per VRF).
        // +optional
+       // +listType=map
+       // +listMapKey=vrf
        Routers []Router `json:"routers"`
        // BFDProfiles is the list of bfd profiles to be used when configuring the neighbors.
        // +optional
@@ -73,8 +75,11 @@ type Router struct {
        ID string `json:"id,omitempty"`
        // VRF is the host vrf used to establish sessions from this router.
        // +optional
-       VRF string `json:"vrf,omitempty"`
+       // +kubebuilder:default:=""
+       VRF string `json:"vrf"`
        // Neighbors is the list of neighbors we want to establish BGP sessions with.
+       // +listType=map
+       // +listMapKey=address
        // +optional
        Neighbors []Neighbor `json:"neighbors,omitempty"`
        // Prefixes is the list of prefixes we want to advertise from this router instance.
@@ -82,6 +87,7 @@ type Router struct {
        Prefixes []string `json:"prefixes,omitempty"`
 }

+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.enableGracefulRestart) || has(self.enableGracefulRestart)", message="Value is required once set"
 // Neighbor represents a BGP Neighbor we want FRR to connect to.
 type Neighbor struct {
        // ASN is the AS number to use for the local end of the session.
@@ -96,6 +102,7 @@ type Neighbor struct {
        SourceAddress string `json:"sourceaddress,omitempty"`

        // Address is the IP address to establish the session with.
+       //+kubebuilder:validation:Required
        Address string `json:"address"`

        // Port is the port to dial when establishing the session.
@@ -146,6 +153,7 @@ type Neighbor struct {
        // EnableGracefulRestart allows BGP peer to continue to forward data packets along
        // known routes while the routing protocol information is being restored.
        // This field is immutable because it requires restart of the BGP session
+       // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
        // +optional
        EnableGracefulRestart bool `json:"enableGracefulRestart,omitempty"`
```

internal/webhooks/frrconfiguration_webhook.go
```diff
@@ -67,24 +67,29 @@ func (v *FRRConfigValidator) Handle(ctx context.Context, req admission.Request)
...
 	case v1.Delete:
-		err := validateConfigDelete(&config)
+		w, err := validateConfigDelete(&config)
 		if err != nil {
 			return admission.Denied(err.Error())
 		}
+		warnings = w
 	}
-	return admission.Allowed("")
+	fmt.Printf("\n%+v\n", warnings)
```
fedepaol, line 91: this is a leftover?

internal/webhooks/frrconfiguration_webhook.go
```diff
@@ -151,6 +158,21 @@ func validateConfig(frrConfig *v1beta1.FRRConfiguration) error {
 				// shouldn't happen for creates as it would be considered an update, and in any case
 				// we add the updated one at the end because for updates we don't want the old and updated resources
 				// to be considered together.
+				for _, rold := range cfg.Spec.BGP.Routers {
+					for _, rnew := range frrConfig.Spec.BGP.Routers {
+						if rold.ID != rnew.ID {
```
fedepaol, line 163: ID might be empty, we should use VRF here

internal/webhooks/frrconfiguration_webhook.go
```diff
@@ -151,6 +158,21 @@ func validateConfig(frrConfig *v1beta1.FRRConfiguration) error {
...
+							continue
+						}
+						for _, nold := range rold.Neighbors {
+							for _, nnew := range rnew.Neighbors {
+								// router bgp 7003' -c 'neighbor 11.11.11.1 graceful-restart'
+								if nold.ASN == nnew.ASN && nold.Address == nnew.Address && nold.EnableGracefulRestart != nnew.EnableGracefulRestart {
+									warnings = append(warnings, "Graceful restart configuration changed, reset this peer to take effect")
```
fedepaol, line 170: you can continue here

internal/webhooks/frrconfiguration_webhook.go
```diff
@@ -151,6 +158,21 @@ func validateConfig(frrConfig *v1beta1.FRRConfiguration) error {
...
 				// we add the updated one at the end because for updates we don't want the old and updated resources
 				// to be considered together.
+				for _, rold := range cfg.Spec.BGP.Routers {
+					for _, rnew := range frrConfig.Spec.BGP.Routers {
+						if rold.ID != rnew.ID {
+							continue
+						}
+						for _, nold := range rold.Neighbors {
+							for _, nnew := range rnew.Neighbors {
+								// router bgp 7003' -c 'neighbor 11.11.11.1 graceful-restart'
```
fedepaol, line 168: is this a leftover?

internal/webhooks/frrconfiguration_webhook.go
fedepaol: also, we don't offer a way to reset, I'd mention something like "it will be available on the next restart" too

internal/webhooks/frrconfiguration_webhook.go
karampok: yes, done

internal/webhooks/frrconfiguration_webhook.go
karampok: done

internal/webhooks/frrconfiguration_webhook.go
karampok: not really, this was on purpose for discussion. Given the config command has asn, ip and gr those should be the two things I compare. Do you agree? ( I removed the comment)

internal/webhooks/frrconfiguration_webhook.go
karampok: done

karampok (commented):
changes in the throw warning commits, thanks

api/v1beta1/frrconfiguration_types.go
```diff
@@ -143,8 +143,9 @@ type Neighbor struct {
 	BFDProfile string `json:"bfdProfile,omitempty"`
 
 	// EnableGracefulRestart allows BGP peer to continue to forward data packets along
-	// known routes while the routing protocol information is being restored.
-	// This field is immutable because it requires restart of the BGP session
+	// known routes while the routing protocol information is being restored.  If
```
fedepaol, line 146: I'd reword the warning with: "if the session is already established, the configuration will have effect after reconnecting to the peer."

e2etests/tests/session.go
```diff
@@ -178,6 +178,84 @@ var _ = ginkgo.Describe("Advertisement", func() {
 			for _, c := range frrs {
 				ValidateFRRPeeredWithNodes(nodes, c, family)
 			}
+
+		},
+			ginkgo.Entry("IPV4", ipfamily.IPv4),
+			ginkgo.Entry("IPV6", ipfamily.IPv6),
+		)
+
+		ginkgo.DescribeTable("Establishes sessions with graceful restart", func(family ipfamily.Family) {
```
fedepaol, line 187: I'd merge this and the other e2e test commit

e2etests/tests/gracefulrestart.go
```diff
@@ -0,0 +1,128 @@
...
+						return fmt.Errorf("Neigh %s does not have prefix %s", p.FRR.Name, prefix)
+					}
+				}
+				return nil
+			}
+
+			Eventually(check, time.Minute, time.Second).ShouldNot(HaveOccurred(),
+				"route should exist before we restart speaker")
+
+			go func() { // go restart speakers while Consistently check that route exist
```
fedepaol, line 111: Why this is different from how we implemented in MetalLB? As per the other discussion, I'd try to avoid spinning go routines. Here we don't know how long it will take, if it's gonna drag up to other tests.

If we have to, we must ensure the go routine ends before we move to the next test.

e2etests/tests/gracefulrestart.go
```diff
@@ -0,0 +1,128 @@
...
+				defer ginkgo.GinkgoRecover()
+				err := cs.CoreV1().Pods(k8s.FRRK8sNamespace).DeleteCollection(context.Background(), metav1.DeleteOptions{}, metav1.ListOptions{
+					LabelSelector: k8s.FRRK8sLabelSelector,
+				})
+				Expect(err).NotTo(HaveOccurred())
+				//DISCUSS: should we trust the delete happens or should we have logic to verify it?
```
fedepaol, line 117: Not only we must ensure the pods got deleted, we also need to check that the new instances are up before we move up to the next test.

e2etests/tests/gracefulrestart.go
```diff
@@ -0,0 +1,128 @@
...
+					Expect(err).NotTo(HaveOccurred())
+					if !found {
+						return fmt.Errorf("Neigh %s does not have prefix %s", p.FRR.Name, prefix)
+					}
+				}
+				return nil
+			}
+
+			Eventually(check, time.Minute, time.Second).ShouldNot(HaveOccurred(),
+				"route should exist before we restart speaker")
```
fedepaol, line 109: nit: frr-k8s instance instead of speaker

e2etests/tests/gracefulrestart.go
fedepaol, line 111: same, we don't have speakers here

e2etests/tests/gracefulrestart.go
fedepaol, line 1: graceful_restart.go

e2etests/tests/gracefulrestart.go
karampok: done

e2etests/tests/gracefulrestart.go
karampok: Can we elaborate more:
what scenario the delete returns without error but pods are not actually deleted?

why we "need to check new instances are up" is that because other test will fail or you think is important for this test?

e2etests/tests/gracefulrestart.go
karampok: done

e2etests/tests/gracefulrestart.go
karampok: Starting from clean state this is the straightfoward implementation imo. 
Serializing the delete and make instantly the monitoring cannot be safe (e.g.  pod delete, routes down, pod up, route up happens before the docker exec vtysh -c show routes is being able to be executed)

(done the wait go routine to end)

karampok: done

e2etests/tests/session.go
karampok:
````
* e53addc - (HEAD -> graceful-restart, karampok/graceful-restart) E2E: add integration test for Graceful Restart (3 hours ago) <karampok>
* 439ef66 - Throw warning when graceful restart flag changes (4 days ago) <karampok>
* 56e8b42 - E2E: correct ginkgo describe name for session.go file (7 days ago) <karampok>
* 9285c6c - E2E: extend tests to check graceful restart (7 days ago) <karampok>
* 6d571ed - API: add graceful restart support (7 days ago) <karampok>
````

Do you want at the end to have only two commits?

File: api/v1beta1/frrconfiguration_types.go (hunk 7 above)
Comment (karampok) [reply api/v1beta1/frrconfiguration_types.go]: done

File: e2etests/pkg/k8s/pods.go [hunk 12]
```diff
@@ -32,3 +40,70 @@ func FRRK8sPods(cs clientset.Interface) ([]*corev1.Pod, error) {
...
+		}
+	}
+
+	return true
+}
+
+func FRRK8isDaemonSetReady(clientset clientset.Interface) (bool, error) {
+
+	// common labels in heml, kustomize
+	l := "app.kubernetes.io/component=frr-k8s,app.kubernetes.io/name=frr-k8s"
```
karampok, line 93: A bit uncertain if worth to put as const and on top of a file

API-DOCS.md
fedepaol, line 1: Not sure you squashed too much but, nit about the commit subj / body:
It's not only about api but about frr-k8s as a whole, so I'd rename it to "Add graceful restart support" or just "Graceful restart support". Also, the commit body is only about the webhook.

api/v1beta1/frrconfiguration_types.go
fedepaol: There's something odd, I still see the old description here as comment but hte new one in the generated crds (while it should be the same, I think?)

api/v1beta1/frrconfiguration_types.go
```diff
@@ -143,9 +143,9 @@ type Neighbor struct {
 	BFDProfile string `json:"bfdProfile,omitempty"`
 
 	// EnableGracefulRestart allows BGP peer to continue to forward data packets along
-	// known routes while the routing protocol information is being restored.  If
-	// EnableGracefulRestart configuration changes, this BGP peer must be reset
-	// to take effect
+	// known routes while the routing protocol information is being restored. If
+	// the session is already established, the configuration will have effect
```
fedepaol, line 147: nit: belongs to the previous commit

e2etests/e2etest_suite_test.go
```diff
@@ -65,6 +67,8 @@ func TestE2E(t *testing.T) {
 }
 
 var _ = ginkgo.BeforeSuite(func() {
+	log.SetLogger(zap.New(zap.WriteTo(ginkgo.GinkgoWriter), zap.UseDevMode(true)))
```
fedepaol, line 70: can you move this to a separate commit?

e2etests/pkg/config/from_containers.go
```diff
@@ -25,33 +32,57 @@ type PeersConfig struct {
...
 	for _, f := range frrs {
-		addresses := f.AddressesForFamily(ipFam)
+
 		ebgpMultihop := false
 		if f.NeighborConfig.MultiHop && f.NeighborConfig.ASN != f.RouterConfig.ASN {
 			ebgpMultihop = true
 		}
 
-		for _, address := range addresses {
+		for _, address := range f.AddressesForFamily(ipFam) {
```
fedepaol, line 69: I think this is making it less readeable

e2etests/pkg/config/from_containers.go
```diff
@@ -25,33 +32,57 @@ type PeersConfig struct {
 	Secrets []corev1.Secret
 }
 
+func (pc PeersConfig) Peers() []Peer {
+	return append(pc.PeersV4, pc.PeersV6...)
+}
+
+func (pc PeersConfig) FRRK8sNeibghbor() []frrk8sv1beta1.Neighbor {
```
fedepaol, line 39: typo: Neibghor

e2etests/pkg/k8s/pods.go
```diff
@@ -32,3 +40,70 @@ func FRRK8sPods(cs clientset.Interface) ([]*corev1.Pod, error) {
...
+			return false
+		}
+	}
+
+	return true
+}
+
+func FRRK8isDaemonSetReady(clientset clientset.Interface) (bool, error) {
+
+	// common labels in heml, kustomize
```
fedepaol, line 92: typo: heml

e2etests/tests/session.go
fedepaol: we said it was fine not to :P

e2etests/pkg/config/from_containers.go
fedepaol: also, it should be Neighbors, and also I think this can be a function and not a method. Something like `NeighborsFrom`.

e2etests/pkg/config/from_containers.go
```diff
@@ -25,33 +32,57 @@ type PeersConfig struct {
...
+	}
+	return res
+}
+
 // PeersForContainers returns two lists of Peers, one for v4 addresses and one for v6 addresses.
 func PeersForContainers(frrs []*frrcontainer.FRR, ipFam ipfamily.Family) PeersConfig {
+	return PeersForContainersWithOptions(frrs, ipFam, AllowAllDisabled, GracefulRestartDisabled)
+}
+
+func PeersForContainersWithOptions(frrs []*frrcontainer.FRR, ipFam ipfamily.Family, allowAll, gracefulRestart bool) PeersConfig {
```
fedepaol, line 55: I'd use the functional options pattern here  https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

e2etests/pkg/k8s/pods.go
fedepaol: let's use the local one, we are using it only here after all

e2etests/pkg/k8s/pods.go
```diff
@@ -32,3 +40,70 @@ func FRRK8sPods(cs clientset.Interface) ([]*corev1.Pod, error) {
...
+	}
+
+	return wait.PollUntilContextTimeout(context.Background(), 100*time.Millisecond,
+		1*time.Minute, false, func(context.Context) (bool, error) {
+			npods, err := FRRK8sPods(cs)
+			if err != nil {
+				return false, err
+			}
+			for _, p := range npods {
+				if slices.Contains(oldNames, p.Name) {
```
fedepaol, line 65: we should also check that the number of the new pods matches the old one.

e2etests/tests/graceful_restart.go
```diff
@@ -0,0 +1,140 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		err = cleanup(updater)
+		Expect(err).NotTo(HaveOccurred(), "cleanup config in API and infra containers")
+
+		cs = k8sclient.New()
+		nodes, err = k8s.Nodes(cs)
+		Expect(err).NotTo(HaveOccurred())
+
+		h, err := k8s.FRRK8isDaemonSetReady(cs)
```
fedepaol, line 63: This is probably redundant, I'd give the frrk8s pod readiness for granted when we run each test.

fedepaol: then let's have it in the beforeSuite, but not before just one test out of many

e2etests/pkg/config/from_containers.go
fedepaol: you can have one function that takes the generated PeersConfig and iterates applying the changes. Or you can have a neighborModifier. I don't like the `WithOptions` wrapper that is `WithOptionsThatINeedOnlyForThisTest`.

e2etests/e2etest_suite_test.go
fedepaol: yes I am not saying we shouldn't add it, just that it doesn't belong here

e2etests/tests/graceful_restart.go
karampok: done

e2etests/pkg/config/from_containers.go
karampok: done

e2etests/pkg/config/from_containers.go
```diff
@@ -77,9 +81,38 @@ func PeersForContainers(frrs []*frrcontainer.FRR, ipFam ipfamily.Family) PeersCo
...
+	}
 	return res
 }
 
+func EnableGracefulRestart(pc *PeersConfig) {
+	// for _,p := range pc.PeersV4 { // not working, go uses a copy of the value instead of the value itself within a range clause
+	//    p.Neigh.EnableGracefulRestart = true
+	// }
+
+	f := func(t []Peer) {
```
fedepaol, line 96: let's just do the for

e2etests/pkg/config/from_containers.go
```diff
@@ -77,9 +81,38 @@ func PeersForContainers(frrs []*frrcontainer.FRR, ipFam ipfamily.Family) PeersCo
...
+}
+
+func EnableAllowAll(pc *PeersConfig) {
+	f := func(t []Peer) {
+		for i := 0; i < len(t); i++ {
+			t[i].Neigh.ToAdvertise.Allowed.Mode = frrk8sv1beta1.AllowAll
+		}
+	}
+
+	f(pc.PeersV4)
```
fedepaol, line 113: same

e2etests/tests/graceful_restart.go
```diff
@@ -0,0 +1,140 @@
...
+		if ginkgo.CurrentSpecReport().Failed() {
+			testName := ginkgo.CurrentSpecReport().LeafNodeText
+			dump.K8sInfo(testName, reporter)
+			dump.BGPInfo(testName, infra.FRRContainers, cs)
+		}
+	})
+
+	ginkgo.Context("When restarting the frrk8s deamon pods", func() {
+
+		assertFuncPerEntry := func(ipFam ipfamily.Family, prefix string) {
```
fedepaol, line 75: not blocking, leaving only because we might change the other comments, but: can we embed this as part of the table entry? It's gonna be as per all the other table tests and we are not really reusing the function

fedepaol:
couple of nits and it's good to go, thanks!

e2etests/pkg/config/from_containers.go
karampok: done

e2etests/pkg/config/from_containers.go
karampok: done

e2etests/tests/graceful_restart.go
karampok: done (but imo is less readable that way)

fedepaol:
sending to merge queue, thanks!

PR #167, opened 2024-07-05
Export the type safe - code-gen generated clients

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Since we want other controllers to consume the frr-k8s api, here we:

- remove the dependencies added by the webhook integration under api, so importing will bring less dependencies
- use a local copy of secret reference instead of the one exported by the k8s api, to reduce the dependency scope
- generate the type safe clientset / listers / informers to be consumed

The introduction of the local `core` symlink is to circumvent the limitation listed in https://github.com/kubernetes/code-generator/issues/167

**Release note**:

```release-note
Make the type safe code-generator generated types available.
```


oribon (approved):
lgtm

PR #168, opened 2024-07-11
Make the frr generated files compatible with 8.0+

fedepaol:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

FRR 8 preferred format for bfd echo-intervals is with separate transmit
/ receive timers. FRR 9.1 actually reports an error when deleting the
echo interval stanza.

Here we:
- generate an frr 8.0+ compatible frr configuration file
- declare the format of the configuration file to be 8.0 compatible

**Release note**:

```release-note
Make the generated frr file compatible with frr 8+: split bfd echo interval in echo tx / echo rx.
```

### DIFF
--- a/internal/frr/templates/bfdprofile.tmpl
+++ b/internal/frr/templates/bfdprofile.tmpl
@@ -13,7 +13,8 @@
     echo-mode
     {{end -}}
     {{ if .profile.EchoInterval -}}
-    echo-interval {{.profile.EchoInterval}}
+    echo transmit-interval {{.profile.EchoInterval}}
+    echo receive-interval {{.profile.EchoInterval}}
     {{end -}}
     {{ if .profile.PassiveMode -}}
     passive-mode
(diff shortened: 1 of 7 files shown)


internal/frr/templates/bfdprofile.tmpl
```diff
@@ -13,12 +13,13 @@
     echo-mode
     {{end -}}
     {{ if .profile.EchoInterval -}}
-    echo-interval {{.profile.EchoInterval}}
+    echo transmit-interval {{.profile.EchoInterval}}
+    echo receive-interval {{.profile.EchoInterval}}
```
karampok, line 17: This should ideally be different commit?

karampok (approved):

internal/frr/templates/bfdprofile.tmpl
fedepaol: nope, it's actually the core of the commit "make the generated files compatible to 8.0+".
From 8.0 frr format has separate echo-intervals, to the point that frr 9+ is throwing an error when deleting echo-interval (the ci failure I got in metallb).

Let me modify the description of the commit to make it clearer.

karampok (approved):

PR #194, opened 2024-08-29
Add support for "Dynamic ASN"

oribon:

/kind feature

Adds support for FRR's "Dynamic ASN", allowing users to specify whether the peer is ibgp or ebgp without explicitly setting its asn. Leverages https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighbor-PEER-remote-as-internal

**Release note**:

```release-note
Add DynamicASN field for a neighbor, which allows the daemon to detect the AS number to use without explicitly setting it. The new field is mutually exclusive with the existing ASN field, and one of them must be specified for any given Neighbor. 
```

### DIFF
--- a/api/v1beta1/frrconfiguration_types.go
+++ b/api/v1beta1/frrconfiguration_types.go
@@ -95,9 +95,20 @@ type Import struct {
 // Neighbor represents a BGP Neighbor we want FRR to connect to.
 type Neighbor struct {
 	// ASN is the AS number to use for the local end of the session.
+	// ASN and DynamicASN are mutually exclusive and one of them must be specified.
 	// +kubebuilder:validation:Minimum=0
 	// +kubebuilder:validation:Maximum=4294967295
-	ASN uint32 `json:"asn"`
+	// +optional
+	ASN uint32 `json:"asn,omitempty"`
+
+	// DynamicASN detects the AS number to use for the local end of the session
+	// without explicitly setting it via the ASN field. Limited to:
+	// internal - if the neighbor's ASN is different than the router's the connection is denied.
+	// external - if the neighbor's ASN is the same as the router's the connection is denied.
+	// ASN and DynamicASN are mutually exclusive and one of them must be specified.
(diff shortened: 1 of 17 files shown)


api/v1beta1/frrconfiguration_types.go
```diff
@@ -95,9 +95,20 @@ type Import struct {
...
-	ASN uint32 `json:"asn"`
+	// +optional
+	ASN uint32 `json:"asn,omitempty"`
+
+	// DynamicASN detects the AS number to use for the local end of the session
+	// without explicitly setting it via the ASN field. Limited to:
+	// internal - if the neighbor's ASN is different than the router's the connection is denied.
+	// external - if the neighbor's ASN is the same as the router's the connection is denied.
+	// ASN and DynamicASN are mutually exclusive and one of them must be specified.
+	// +kubebuilder:validation:Enum=internal;external
```
fedepaol, line 109: Can we enforce the mutual exclusivity (or -ness) with CEL rules?
https://kubernetes.io/docs/reference/using-api/cel/

internal/frr/config.go
```diff
@@ -187,11 +187,28 @@ func templateConfig(data interface{}) (string, error) {
...
+				// return true only for non-multihop IPv6 eBGP sessions
+				if ipFamily == ipfamily.IPv4 {
+					return false
+				}
+
+				if eBGPMultiHop {
+					return false
+				}
+
+				if asn == "internal" {
```
fedepaol, line 200: let's either have a comment saying that all these ifs mean it's external (or wrap in a "isEBGP" helper)

oribon: done

api/v1beta1/frrconfiguration_types.go
oribon: tried for a while, did not go so well 😅  I guess we can tackle it in a different time (applies for password and passwordSecret as well)

fedepaol:
lgtm

PR #210, opened 2024-10-17
E2E: slim the graceful restart test

karampok:
This e2e test flakes often, and especially when the github actions
are slow. Therefore we reduce the overall peering to make the test
require less resources. What this commit does:
1. GR test will only use peering from single node to ebgp-single-hop
(so we will run 4x less `docker exec` to grab the routes).
2. GR test will restart single frr-k8s pod and will monitor that single
route and wait only that pod to be healthy.
3. Given we target the GR between specific host and specific peer, we
know which logs to look in case of failure.

 /kind flake

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/pkg/config/from_containers.go
+++ b/e2etests/pkg/config/from_containers.go
@@ -3,6 +3,7 @@
 package config
 
 import (
+	"fmt"
 	"net"
 
 	frrk8sv1beta1 "github.com/metallb/frr-k8s/api/v1beta1"
@@ -135,4 +136,13 @@ func ContainersForVRF(frrs []*frrcontainer.FRR, vrf string) []*frrcontainer.FRR 
 		}
 	}
 	return res
+}
+
+func ContainerByName(frrs []*frrcontainer.FRR, name string) (*frrcontainer.FRR, error) {
+	for _, f := range frrs {
+		if f.Name == name {
+			return f, nil
+		}
+	}
+	return nil, fmt.Errorf("container with name %s not found", name)
 }
--- a/e2etests/pkg/k8s/pods.go
+++ b/e2etests/pkg/k8s/pods.go
@@ -5,12 +5,10 @@ package k8s
 import (
 	"context"
 	"fmt"
-	"slices"
 	"time"
 
 	"errors"
 
-	"github.com/onsi/ginkgo/v2"
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -42,40 +40,45 @@ func FRRK8sPods(cs clientset.Interface) ([]*corev1.Pod, error) {
 	return res, nil
 }
 
-func RestartFRRK8sPods(cs clientset.Interface) error {
-	pods, err := FRRK8sPods(cs)
+func RestartFRRK8sPodForNode(cs clientset.Interface, nodeName string) error {
+	oldPod, err := podForNode(cs, nodeName)
 	if err != nil {
 		return err
 	}
-	oldNames := []string{}
-	for _, p := range pods {
-		oldNames = append(oldNames, p.Name)
-		err := cs.CoreV1().Pods(FRRK8sNamespace).Delete(context.Background(), p.Name, metav1.DeleteOptions{})
-		if err != nil {
-			return err
-		}
+	if err := cs.CoreV1().Pods(FRRK8sNamespace).Delete(context.Background(), oldPod.Name, metav1.DeleteOptions{}); err != nil {
+		return err
 	}
 
 	return wait.PollUntilContextTimeout(context.Background(), 10*time.Second,
 		120*time.Second, false, func(context.Context) (bool, error) {
-			npods, err := FRRK8sPods(cs)
+			newPod, err := podForNode(cs, nodeName)
 			if err != nil {
 				return false, err
 			}
-			if len(npods) != len(pods) {
+
+			if newPod.Name == oldPod.Name {
+				return false, fmt.Errorf("pod was not deleted")
+			}
+
+			if !podIsRunningAndReady(newPod) {
 				return false, nil
 			}
-			for _, p := range npods {
-				if slices.Contains(oldNames, p.Name) {
-					return false, nil
-				}
-				if !podIsRunningAndReady(p) {
-					ginkgo.By(fmt.Sprintf("\t%v pod %s not ready", time.Now(), p.Name))
-					return false, nil
-				}
-			}
+
 			return true, nil
 		})
+}
+
+func podForNode(cs clientset.Interface, nodeName string) (*corev1.Pod, error) {
+	pods, err := FRRK8sPods(cs)
+	if err != nil {
+		return nil, err
+	}
+	for _, p := range pods {
+		if p.Spec.NodeName == nodeName {
+			return p, nil
+		}
+	}
+	return nil, fmt.Errorf("no frr-k8s pod found on node %s", nodeName)
 }
 
 func podIsRunningAndReady(pod *v1.Pod) bool {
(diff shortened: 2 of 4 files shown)


karampok:
As part of the testing I have run the test 10 times (https://github.com/metallb/frr-k8s/actions/runs/11381447010?pr=210) and observed not failure.

karampok:
@fedepaol the failing test seems not related. Ready for feedback, thanks

e2etests/tests/graceful_restart.go
```diff
@@ -68,12 +68,15 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
...
+
+		err := cleanup(updater)
+		Expect(err).NotTo(HaveOccurred(), "cleanup config in API and in infra containers")
 	})
 
 	ginkgo.Context("When restarting the frrk8s deamon pods", func() {
 
 		ginkgo.DescribeTable("external BGP peer maintains routes", func(ipFam ipfamily.Family, prefix string) {
-			frrs := config.ContainersForVRF(infra.FRRContainers, "")
+			frrs := config.ContainersForName(infra.FRRContainers, "ebgp-single-hop")
```
fedepaol, line 79: if it's a single container, no need for the for loop below

fedepaol:
Can you explain the rationale on why this improves the stability of the test in the commit message? Here we are declaring only what we are doing, not why

e2etests/pkg/k8s/pods.go
```diff
@@ -42,13 +40,16 @@ func FRRK8sPods(cs clientset.Interface) ([]*corev1.Pod, error) {
 	return res, nil
 }
 
-func RestartFRRK8sPods(cs clientset.Interface) error {
+func RestartFRRK8sPods(cs clientset.Interface, node string) error {
```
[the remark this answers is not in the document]
fedepaol: Ignore the wrapper part. I thought we were using this beyond the gr tests.

I'd like this to do less. So let's have one function for `RestartFRRK8sPod(cs, pod)` and a way to retrieve the pod running on a given node (if we don't have it already).

From outside it's going to be

```
pod := podForNode(node)
restartPod(pod)
```

which I think is more straightforward. Also, given that we are talking about one single pod here, the waiting logic below can simplified.

e2etests/pkg/config/from_containers.go
```diff
@@ -136,3 +136,14 @@ func ContainersForVRF(frrs []*frrcontainer.FRR, vrf string) []*frrcontainer.FRR
 	}
 	return res
 }
+
+func ContainersForName(frrs []*frrcontainer.FRR, name string) []*frrcontainer.FRR {
```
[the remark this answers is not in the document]
karampok: done

e2etests/pkg/k8s/pods.go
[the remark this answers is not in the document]
karampok: I have removed the not need loop logic, nevertheless, everything is in the same function because I need the node name to get the pod(oldPod), then delete the pod, then need the node name to fetch the pod (newPod), compare old,new are diff and status.

e2etests/pkg/k8s/pods.go
```diff
@@ -42,38 +40,43 @@ func FRRK8sPods(cs clientset.Interface) ([]*corev1.Pod, error) {
 	return res, nil
 }
 
-func RestartFRRK8sPods(cs clientset.Interface) error {
-	pods, err := FRRK8sPods(cs)
+func RestartFRRK8sPod(cs clientset.Interface, nodeName string) error {
+	podForNode := func() (*corev1.Pod, error) {
```
fedepaol, line 44: let's have this defined outside (below)

e2etests/pkg/k8s/pods.go
```diff
@@ -42,38 +40,43 @@ func FRRK8sPods(cs clientset.Interface) ([]*corev1.Pod, error) {
 	return res, nil
 }
 
-func RestartFRRK8sPods(cs clientset.Interface) error {
-	pods, err := FRRK8sPods(cs)
+func RestartFRRK8sPod(cs clientset.Interface, nodeName string) error {
```
fedepaol, line 43: Let's rename this "RestartFRRK8sPodForNode".

e2etests/pkg/k8s/pods.go
karampok: done

karampok:
The  `metallb_e2e` fails again same
```
FAILED] Timed out after 120.000s.
  Unexpected error:
      <*errors.errorString | 0xc00045d680>: 
      metric frrk8s_bgp_session_up not found
      {
          s: "metric frrk8s_bgp_session_up not found",
      }
  occurred
```
is it  known issue?

fedepaol:
> This e2e test flakes often, and especially when the github actions are slow. Therefore we reduce the overall peering to make the test require less resources. What this commit does:
> 
>     1. GR test will only use peering from single node to ebgp-single-hop
>        (so we will run 4x less `docker exec` to grab the routes).
> 

Is this a real time consuming issue?

>     2. GR test will restart single frr-k8s pod and will monitor that single
>        route and wait only that pod to be healthy.
> 
>     3. Given we target the GR between specific host and specific peer, we
>        know which logs to look in case of failure.
> 

This shouldn't be a blocker, if we have the proper failure information.

> 

All in all, I think we should cover all the 4 combinations (i/e bgp - single / multi hop). The easiest is to iterate over the containers and run this test (in this PR's version) only once. Optimizations are, trying to understand if that docker exec is really an offender and in case see if we can spin up go routines to make that faster.

If I had to bet, restarting a pod takes a lot of time but collecting routes should not. WDYT?

fedepaol:
> The `metallb_e2e` fails again same
> 
> ```
> FAILED] Timed out after 120.000s.
>   Unexpected error:
>       <*errors.errorString | 0xc00045d680>: 
>       metric frrk8s_bgp_session_up not found
>       {
>           s: "metric frrk8s_bgp_session_up not found",
>       }
>   occurred
> ```
> 
> is it known issue?

could be related to the change of the port in metrics https://github.com/metallb/frr-k8s/pull/206
I don't remember if I overrode the test or if it passed in that pr. Will check, but it's probably independent from this. I'll file a dummy pr to ensure it

karampok:
> Is this a real time consuming issue?
I think yes, but if run the test one after another like: for ibgp-single peer, restart the pod, routes ok, pod ok, then start again for ebgp  then the spike on the system resources will be lower (in a period of 2minutes, I do 120 times docker exec time number of container)

> If I had to bet, restarting a pod takes a lot of time but collecting routes should not. WDYT?

I would say is 70% pod takes long time, 30% docker exec fails. The diff here the route check (docker exec return routes) must always succeed, single failure fails the test. No other test require that granularity (eventually it just retries and temp failures are hidden).

We do need the get routes happening as fast as possible and the restart pod in the background to observe route removal (bgp graceful-restart preserve-fw-state)

Also to be in safe side 2m or 1m is hardcoded, because I cannot stop the test earlier once the pod is ready.


One suggestion is that I only reduce the restart pod to one node in the PR, and we observe how much flaky the test is before we refactor.

karampok:
Another idea would be to change the checking logic and instead of checking routes exists that I will check `show ip route` which includes the age of the route and if age is renewed to fail.


```
frr-k8s-control-plane# show ip route
Codes: K - kernel route, C - connected, S - static, R - RIP,
       O - OSPF, I - IS-IS, B - BGP, E - EIGRP, N - NHRP,
       T - Table, v - VNC, V - VNC-Direct, A - Babel, F - PBR,
       f - OpenFabric,
       > - selected route, * - FIB route, q - queued, r - rejected, b - backup
       t - trapped, o - offload failure

K>* 0.0.0.0/0 [0/0] via 172.18.0.1, eth0, 16:23:46
C * 10.244.0.1/32 is directly connected, veth6f5e688f, 16:23:46
```
not sure if leverage this in another test

karampok:
https://github.com/metallb/frr-k8s/pull/218 should improve the the test not to fail.

PR #218, opened 2024-10-28
Add route-map delay timer

karampok:
When the default value of 5 seconds is being used, Graceful-Restart can break momentarily on reconnect of BGP peering.

The restarter FRR when that timeout expires and the route-maps have not been processed, denies the routes, and therefore are removed.

```
DEBUG #bgpd[45]: Route-map: null, prefix: 99.0.0.0/24, result: deny
DEBUG #bgpd[45]: 10.0.64.1 [Update:SEND] 99.0.0.0/24 is filtered by route-map
```

https://github.com/FRRouting/frr/pull/10757
https://github.com/FRRouting/frr/issues/10756

/kind bug

**Release note**:

```release-note
Add route-map delay timer to avoid graceful restart breaking due to timing.
```

### DIFF
--- a/internal/frr/templates/frr.tmpl
+++ b/internal/frr/templates/frr.tmpl
@@ -19,6 +19,8 @@ debug route-map detail
 hostname {{.Hostname}}
 ip nht resolve-via-default
 ipv6 nht resolve-via-default
+bgp route-map delay-timer 20
+
 
 {{- range $r := .Routers }}
 {{- range .Neighbors }}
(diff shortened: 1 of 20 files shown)


karampok:
Given that CI is green I would say it is okay to add that config, what do you think?  

We could potentially use it only when graceful restart is enabled but that will take a bit of refactoring (is not per neighbor)

fedepaol:
> Given that CI is green I would say it is okay to add that config, what do you think?
> 

I am not sure I understand. The linked issue https://github.com/FRRouting/frr/issues/10756 is merged and backported to (at least) 8.2 while here we use 9.0. Are you sure we are hitting that bug? Does it mean that frr is affected by https://github.com/FRRouting/frr/issues/10756  ?

karampok:
AFAIU we do not hit the bug, this bug kind of desribes there is a route-map delay timer which I was not aware neither I can find documentation about. Tuning the timer to the value 0 reproduces the failure instantly, tuning the value to 20 does not fail for 150+ times repeating the test. So doing GR in our ci (maybe in the field) and having 5 seconds might give a bug.

PR #221, opened 2024-11-05
Throw better error when prefix for same localPref is added twice

karampok:
Currently the return error for duplicate prefix
```
 withLocalPref:
  - prefixes:
      - "4.4.4.100/32"
      - "4.4.4.100/32"
    localPref: 100
```
is `multiple local prefs specified for prefix 4.4.4.100/32` which is misleading.

/kind bug

**Release note**:

```release-note
bug: throw better error when prefix for same localPref is added twice
```

### DIFF
--- a/internal/controller/api_to_config.go
+++ b/internal/controller/api_to_config.go
@@ -514,7 +514,17 @@ func localPrefPrefixesToMap(withLocalPref []v1beta1.LocalPrefPrefixes) (localPre
 		localPrefForPrefixV6: map[string]uint32{},
 	}
 
+	seen := make(map[uint32]bool)
 	for _, pfxs := range withLocalPref {
+		if _, exists := seen[pfxs.LocalPref]; exists {
+			// error when input is withLocalPref = []v1beta1.LocalPrefPrefixes{
+			//	                      Prefixes:  []string{"192.0.2.0/24"},LocalPref: 100}, []string{"100.0.0.0/24"},LocalPref: 100}}
+			// so we enforce       withLocalPref = []v1beta1.LocalPrefPrefixes{
+			//	                      Prefixes:  []string{"192.0.2.0/24, "100.0.0.0/24"},LocalPref: 100}}
+			return localPrefPrefixes{}, fmt.Errorf("multiple entries with the same localPref: %d", pfxs.LocalPref)
+		}
+		seen[pfxs.LocalPref] = true
+
 		for _, p := range pfxs.Prefixes {
 			family := ipfamily.ForCIDRString(p)
 			lpMap := res.localPrefForPrefixV4
@@ -522,12 +532,18 @@ func localPrefPrefixesToMap(withLocalPref []v1beta1.LocalPrefPrefixes) (localPre
 				lpMap = res.localPrefForPrefixV6
 			}
 
-			_, ok := lpMap[p]
-			if ok {
-				return localPrefPrefixes{}, fmt.Errorf("multiple local prefs specified for prefix %s", p)
+			v, ok := lpMap[p]
+			if !ok {
+				lpMap[p] = pfxs.LocalPref
+				continue
 			}
 
-			lpMap[p] = pfxs.LocalPref
+			if v != pfxs.LocalPref {
+				return localPrefPrefixes{}, fmt.Errorf("multiple local prefs (%d,%d) specified for prefix %s", v, pfxs.LocalPref, p)
+			}
+			if v == pfxs.LocalPref {
+				return localPrefPrefixes{}, fmt.Errorf("prefix %s with local prefs %d defined twice", p, v)
+			}
 		}
 	}
 
--- a/internal/controller/api_to_config_test.go
+++ b/internal/controller/api_to_config_test.go
@@ -844,6 +844,88 @@ func TestConversion(t *testing.T) {
 			err:      fmt.Errorf("multiple local prefs specified for prefix %s", "192.0.4.0/24"),
 		},
 		{
+			name: "One neighbor, trying to set samelocalPrefs for different prefix entries",
+			fromK8s: []v1beta1.FRRConfiguration{
+				{
+					Spec: v1beta1.FRRConfigurationSpec{
+						BGP: v1beta1.BGPConfig{
+							Routers: []v1beta1.Router{
+								{
+									ASN: 65040,
+									ID:  "192.0.2.20",
+									Neighbors: []v1beta1.Neighbor{
+										{
+											ASN:     65041,
+											Address: "192.0.2.21",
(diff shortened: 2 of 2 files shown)


karampok:
@oribon @fedepaol I think we should  error in that case, but I could see also the alternative to silently allow the case by ignoring same prefix/same local pref entries. What do you prefer?

internal/controller/api_to_config.go
```diff
@@ -522,10 +522,13 @@ func localPrefPrefixesToMap(withLocalPref []v1beta1.LocalPrefPrefixes) (localPre
 				lpMap = res.localPrefForPrefixV6
 			}
 
-			_, ok := lpMap[p]
-			if ok {
+			v, ok := lpMap[p]
+			if ok && v != pfxs.LocalPref {
 				return localPrefPrefixes{}, fmt.Errorf("multiple local prefs specified for prefix %s", p)
 			}
+			if ok && v == pfxs.LocalPref {
```
fedepaol, line 529: I wonder if we should even error here. What happens if we add the same local pref twice (either in the same frrconfiguration or when merging)?

karampok: >  if we add the same local pref twice in the same frrconfiguration
what frrconfig  do you have in mind? I do not see the use case to be honest.

> if we add the same local pref twice when merging
This sounds very legit, then we should not error we should just let it pass and ignore duplicates.
Do you have an example of that use case, is unclear why two frrconfig (two actors) would like to use the same prefix/local pref twice?

fedepaol: > > if we add the same local pref twice in the same frrconfiguration
> 
> what frrconfig do you have in mind? I do not see the use case to be honest.
> 

The end result I expect is having them squashed.

> > if we add the same local pref twice when merging
> 
> This sounds very legit, then we should not error we should just let it pass and ignore duplicates. Do you have an example of that use case, is unclear why two frrconfig (two actors) would like to use the same prefix/local pref twice?

Because they agree the localpref must be that particular value I guess.

oribon: (I think) this function does not know about merges, it works per neighbor. wdyt about enforcing that a single `[]LocalPrefPrefixes` cannot have the same localPref defined multiple times + the same prefix cannot appear in multiple `[]Prefixes` (of the same []LocalPrefPrefixes)?

karampok: @oribon  Can you elaborate a bit more? What would be the error when ` the same prefix cannot appear in multiple []Prefixes (of the same []LocalPrefPrefixes)?`  Should we handle different the following case?
Like error in the first, do not error in the second?

```
PrefixesWithLocalPref: []v1beta1.LocalPrefPrefixes{
{
Prefixes:  []string{"192.0.2.0/24", "192.0.2.0/24"},LocalPref: 100}
}
```

```
PrefixesWithLocalPref: []v1beta1.LocalPrefPrefixes{
{
Prefixes:  []string{"192.0.2.0/24"},LocalPref: 100},
Prefixes:  []string{"192.0.2.0/24"},LocalPref: 100}
},
```

oribon: I'm suggesting to error on both cases (to enforce FRRConfigurations to be cleaner).

internal/controller/api_to_config_test.go
```diff
@@ -843,6 +843,88 @@ func TestConversion(t *testing.T) {
...
+									Prefixes: []string{"192.0.2.0/24"},
+								},
+							},
+						},
+					},
+				},
+			},
+			secrets:  map[string]v1.Secret{},
+			expected: nil,
+			err:      fmt.Errorf("a not nil error"),
```
karampok, line 926: We do not really check the error values, so I found having description here is misleading.

internal/controller/api_to_config.go
```diff
@@ -514,6 +514,18 @@ func localPrefPrefixesToMap(withLocalPref []v1beta1.LocalPrefPrefixes) (localPre
 		localPrefForPrefixV6: map[string]uint32{},
 	}
 
+	// error when input is
+	// Prefixes:  []string{"192.0.2.0/24"},LocalPref: 100},
```
fedepaol, line 518: let's make the case we are addressing clearer here: withLocalPref = []v1beta1.LocalPrefPrefixes{...}

internal/controller/api_to_config.go
```diff
@@ -514,6 +514,18 @@ func localPrefPrefixesToMap(withLocalPref []v1beta1.LocalPrefPrefixes) (localPre
...
+	// Prefixes:  []string{"100.0.0.0/24"},LocalPref: 100}
+	// we enforce : []string{"192.0.2.0/24","100.0.0.0/24"},LocalPref: 100}
+	seen := make(map[uint32]bool)
+	for _, e := range withLocalPref {
+		if _, exists := seen[e.LocalPref]; exists {
+			return localPrefPrefixes{}, fmt.Errorf("duplicate localPref")
```
fedepaol, line 524: let's give a better error here, so it's clear what are the offending elements. Something along the line of "multiple  entries with the same localPref: 100"

fedepaol: Also also, can this be part of the outer loop below? So it's clear that we are addressing different local prefs in each row and we can focus on not having duplicate prefixes: if each row relates to a different local pref, then the only thing we care about in the inner loop is not to have duplicate prefixes

karampok:
thanks for the feedback, updated

internal/controller/api_to_config.go
```diff
@@ -514,18 +514,31 @@ func localPrefPrefixesToMap(withLocalPref []v1beta1.LocalPrefPrefixes) (localPre
...
 		for _, p := range pfxs.Prefixes {
 			family := ipfamily.ForCIDRString(p)
 			lpMap := res.localPrefForPrefixV4
 			if family == ipfamily.IPv6 {
 				lpMap = res.localPrefForPrefixV6
 			}
 
-			_, ok := lpMap[p]
-			if ok {
+			v, ok := lpMap[p]
```
fedepaol, line 535: can we assign and exit early if !ok? This will make the ifs below simpler

internal/controller/api_to_config.go
```diff
@@ -514,18 +514,31 @@ func localPrefPrefixesToMap(withLocalPref []v1beta1.LocalPrefPrefixes) (localPre
...
+			if ok && v != pfxs.LocalPref {
```
fedepaol, line 536: let's add v and the pfxs.LocalPref to the error

fedepaol:
Small nits, then it's good to go. Thanks!

fedepaol:
LGTM

PR #224, opened 2024-11-11
Use metrics for readiness/liveness probes

oribon:
Since the existing healthz/readyz don't do much, we leverage the metrics endpoints for readiness/liveness to avoid opening another port (8081) on the host.

/kind cleanup

**Release note**:

```release-note
NONE
```


fedepaol:
I guess metallb's prometheus rules will need to be changed accordingly.
LGTM

PR #225, opened 2024-11-20
E2E: repeat session open many times

karampok:

/kind bug

> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

In frr-k8s backend (default) we allow the BGP peering to be opened by the peer.
This was not the case with the old backend. I observed that when the external peer is open the connection the BGP peering does not seem to get always the graceful restart capability.

from logs it simply looks the moment the BGP peering is established, the GR config is not place

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -238,7 +238,7 @@ $(APIDOCSGEN): $(LOCALBIN)
 
 .PHONY: e2etests
 e2etests: ginkgo kubectl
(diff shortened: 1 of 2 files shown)


karampok:
not hitting the bug, so close

PR #231, opened 2024-11-26
Add zebra -K argument and extend e2e testing

karampok:
/kind bug

If we use learning routes, when we restart the pod those routes are gone from the k8s node. There is an option for zebra to keep them. That is useful only if Graceful restart is being used.

**Release note**:

```release-note
Add by default the -K (graceful_restart) argument in Zebra so that learnt routes are maintained on the host when zebra process restart.
```

### DIFF
--- a/charts/frr-k8s/templates/controller.yaml
+++ b/charts/frr-k8s/templates/controller.yaml
@@ -49,7 +49,7 @@ data:
     # Check /etc/pam.d/frr if you intend to use "vtysh"!
     #
     vtysh_enable=yes
-    zebra_options="  -A 127.0.0.1 -s 90000000"
+    zebra_options="  -A 127.0.0.1 -K 120 -s 90000000"
     bgpd_options="   -A 127.0.0.1 {{ if not .Values.frrk8s.frr.acceptIncomingBGPConnections }} -p 0 {{- end }}"
     ospfd_options="  -A 127.0.0.1"
     ospf6d_options=" -A ::1"
--- a/e2etests/go.mod
+++ b/e2etests/go.mod
@@ -6,13 +6,13 @@ toolchain go1.22.4
 
 replace (
 	github.com/metallb/frr-k8s => ../
-	go.universe.tf/e2etest => github.com/metallb/metallb/e2etest v0.0.0-20241212081358-ee34c275145f
+	go.universe.tf/e2etest => github.com/metallb/metallb/e2etest v0.0.0-20250127152015-3f9e99adbb0e
 	go.universe.tf/metallb => github.com/metallb/metallb v0.14.5
 )
 
 require (
 	github.com/google/go-cmp v0.6.0
-	github.com/metallb/frr-k8s v0.0.16
+	github.com/metallb/frr-k8s v0.0.17
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
 	github.com/openshift-kni/k8sreporter v1.0.4
--- a/e2etests/tests/graceful_restart.go
+++ b/e2etests/tests/graceful_restart.go
@@ -4,11 +4,13 @@ package tests
 
 import (
 	"fmt"
+	"net/netip"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/openshift-kni/k8sreporter"
-	"go.universe.tf/e2etest/pkg/frr/container"
+	"go.universe.tf/e2etest/pkg/container"
+	"go.universe.tf/e2etest/pkg/executor"
 
 	frrk8sv1beta1 "github.com/metallb/frr-k8s/api/v1beta1"
 	"github.com/metallb/frrk8stests/pkg/config"
@@ -19,6 +21,7 @@ import (
 	"github.com/metallb/frrk8stests/pkg/routes"
 	. "github.com/onsi/gomega"
 	frrconfig "go.universe.tf/e2etest/pkg/frr/config"
+	frrcontainer "go.universe.tf/e2etest/pkg/frr/container"
 	"go.universe.tf/e2etest/pkg/ipfamily"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -68,19 +71,35 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
 			dump.K8sInfo(testName, reporter)
 			dump.BGPInfo(testName, infra.FRRContainers, cs)
 		}
+
+		err := cleanup(updater)
+		Expect(err).NotTo(HaveOccurred(), "cleanup config in API and infra containers")
 	})
 
 	ginkgo.Context("When restarting the frrk8s deamon pods", func() {
-
-		ginkgo.DescribeTable("external BGP peer maintains routes", func(ipFam ipfamily.Family, prefix string) {
+		ginkgo.DescribeTable("both routes on nodes and on the external peers are maintained", func(ipFam ipfamily.Family, prefix string) {
+			externalPrefixes := []string{}
 			frrs := config.ContainersForVRF(infra.FRRContainers, "")
-			for _, c := range frrs {
-				err := container.PairWithNodes(cs, c, ipFam)
+			for i, c := range frrs {
+				l := fmt.Sprintf("200.200.%d.0/24", i)
+				modify := func(frr *frrcontainer.FRR) {
+					frr.NeighborConfig.ToAdvertiseV4 = []string{l}
+				}
+
+				if ipFam == ipfamily.IPv6 {
+					l = fmt.Sprintf("2001:db%d::/64", i)
+					modify = func(frr *frrcontainer.FRR) {
+						frr.NeighborConfig.ToAdvertiseV6 = []string{l}
+					}
+				}
+
+				err := frrcontainer.PairWithNodes(cs, c, ipFam, modify)
 				Expect(err).NotTo(HaveOccurred(), "set frr config in infra containers failed")
+				externalPrefixes = append(externalPrefixes, l)
 			}
 
 			peersConfig := config.PeersForContainers(frrs, ipFam,
-				config.EnableAllowAll, config.EnableGracefulRestart)
+				config.EnableAllowAll, config.EnableReceiveAllowAll, config.EnableGracefulRestart)
 
 			frrConfigCR := frrk8sv1beta1.FRRConfiguration{
 				ObjectMeta: metav1.ObjectMeta{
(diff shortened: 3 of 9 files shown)


karampok:
@fedepaol this is ready for review but probably we need to break into different commits/PR

1. one PR in metallb for adding a test function
2. one commit to add the real change
3. one commit to extent the test

Thanks

fedepaol:
> @fedepaol this is ready for review but probably we need to break into different commits/PR
> 
>     1. one PR in metallb for adding a test function
> 
>     2. one commit to add the real change
> 
>     3. one commit to extent the test
> 
> 
> Thanks

Then please break it :-)

karampok:
sure just confirming this is what you prefer

charts/frr-k8s/templates/controller.yaml
```diff
@@ -49,7 +49,7 @@ data:
     # Check /etc/pam.d/frr if you intend to use "vtysh"!
     #
     vtysh_enable=yes
-    zebra_options="  -A 127.0.0.1 -s 90000000"
+    zebra_options="  -A 127.0.0.1 -K 120 -s 90000000"
```
fedepaol, line 52: the commit says also `--graceful_restart argument ` but I don't see it. 
Also, what's special about 120 seconds? Can we document it (as a comment and in the commit description)?
Also also, would it make sense to have it as a parameter?

e2etests/go.mod
```diff
@@ -6,21 +6,21 @@ toolchain go1.22.4
 
 replace (
 	github.com/metallb/frr-k8s => ../
-	go.universe.tf/e2etest => github.com/metallb/metallb/e2etest v0.0.0-20240715121012-af3c10d65f18
+	go.universe.tf/e2etest => github.com/karampok/metallb/e2etest v0.0.0-20240613071516-2d52178fbc89
```
fedepaol, line 9: to be removed?

e2etests/go.mod
```diff
@@ -6,21 +6,21 @@ toolchain go1.22.4
...
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
 	github.com/openshift-kni/k8sreporter v1.0.4
 	github.com/prometheus/client_model v0.6.1
 	github.com/prometheus/common v0.55.0
 	go.universe.tf/e2etest v0.0.0-00010101000000-000000000000
-	k8s.io/api v0.31.0
-	k8s.io/apimachinery v0.31.0
-	k8s.io/client-go v0.31.0
+	k8s.io/api v0.31.4
```
fedepaol, line 21: are these dep updates relevant?

e2etests/tests/graceful_restart.go
```diff
@@ -128,8 +142,33 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
...
 		},
-			ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.2.0/24"),
-			ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:ccd:e799::/64"),
+			ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.2.0/24", "200.200.200.0/24"),
+			ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:ccd:e799::/64", "2001:db8::/64"),
 		)
 	})
 })
+
+// checkPrefixOnNodes checks that a prefix has at least one route on the
```
fedepaol, line 151: should this belong to some pkg?

e2etests/tests/graceful_restart.go
```diff
@@ -68,19 +72,25 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
 			dump.K8sInfo(testName, reporter)
 			dump.BGPInfo(testName, infra.FRRContainers, cs)
 		}
+
+		err := cleanup(updater)
+		Expect(err).NotTo(HaveOccurred(), "cleanup config in API and infra containers")
 	})
 
 	ginkgo.Context("When restarting the frrk8s deamon pods", func() {
+		ginkgo.DescribeTable("routes are maintained", func(ipFam ipfamily.Family, prefix, learnRoute string) {
```
fedepaol, line 81: can we clarify that we are checking routes both sides? Would be a single test checking each side clearer?

e2etests/go.mod
karampok: Once this https://github.com/metallb/frr-k8s/pull/239 is merged, I would refactor and this PR should not have any changes in the go.mod.

I am drafting the PR until the dependency is merged

e2etests/tests/graceful_restart.go
karampok: Imo this help function is tailored  to the graceful restart scenario because it checks  the routes on the host itself (all other test use the routes from the vtysh) and is relaxed on the next hop (because not all nexthops exist as decribed int he coment). So I would not transfer into a pkg.

No objection though if you want that (which pkg do you have in mind)?

e2etests/tests/graceful_restart.go
karampok: I will extend the description.

> Would be a single test checking each side clearer?

Not sure do you mean two tests? Given that test talkes 2minutes I would not add another instance of it for time reasons. Also checking in two ` consistently ` is not working because I must make sure that routes always exists.

charts/frr-k8s/templates/controller.yaml
karampok: The `-K` is the short version arg for the `--graceful_restart`, I prefered the short, do you prefer to use the long or should improve the commit title?

About the 120seconds, the 120 matches the 120 seconds that graceful-timeout advertise to the other side (https://docs.frrouting.org/en/latest/bgp.html#clicmd-bgp-graceful-restart-restart-time-0-4095) 

I do not think it should be parameter because we already hardcode the https://docs.frrouting.org/en/latest/bgp.html#clicmd-bgp-graceful-restart-restart-time-0-4095 and i do not see any benefit.

karampok:
cc @fedepaol what about this?

e2etests/tests/graceful_restart.go
```diff
@@ -128,8 +142,33 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
...
-			ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.2.0/24"),
-			ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:ccd:e799::/64"),
+			ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.2.0/24", "200.200.200.0/24"),
+			ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:ccd:e799::/64", "2001:db8::/64"),
 		)
 	})
 })
+
+// checkPrefixOnNodes checks that a prefix has at least one route on the
+// each node. There is no check on nexthops because not all routes are install
```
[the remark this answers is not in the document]
karampok: When you have a prefix X, for which there is nexthop from iBGP (via Y) and eBGP (via Z), then the zebra when installing the route in the kernel, it picks one the Z. 

So if you could see routes in vtysh (we cannot we restart the process), you see all nexthops for the prefix, but when we see routes in the kernel we only see the most preferable.

So the test should have
1. if iBGP and eBGP, then check iBGP nexthop
2. if eBGP then pick eBGP

a complexity that not sure is worh it

(plus some different ipv4 vs ipv6 behavior that I need to re-investigate but iirc is about iproute2 return different json outputs)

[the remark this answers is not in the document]
fedepaol: yes, but by doing this we check the presence of the prefix regardless of the fact of who is advertising it. I'd rather have a prefix per peer and make sure all 4 are working properly.
Does this make sense?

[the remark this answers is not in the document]
karampok: Not sure tbh.
That the route learning works in all kind of peers is tested in other tests.
That  a route onced installed by zebra to kernel remains on restarting
the frr-k8d pos is what we test. 
But i see your point, I will work towards that. Thanks for the feedback

(fyi I am going to rebase since that PR is a bit old)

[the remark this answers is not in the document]
fedepaol: Well if that's true, then there's no point in bothering to test with all flavors, and If we don't doubt that the type of peer can affect the behaviour, one external frr would be enough.

Here we are doing a mix, looking at the code we are testing the four flavors (hence my comment of testing them well) while it could be that one flavor is working out and another is not.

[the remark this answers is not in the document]
karampok: I agree with you, since I do not test that in isolation I should check all the flavors. I just pushed a version which has different prefix per peer/flavor. Let me know what you think, thanks

e2etests/tests/graceful_restart.go
```diff
@@ -133,3 +157,27 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
 		)
 	})
 })
+
+// checkPrefixOnNode checks that prefixes have at one kernel route on the
+// the node. This function get routes directly from the node `docker exec ip
+// route` and not from `vtysh show route` because we need to get routes while
+// FRR process restarts.
+func checkPrefixOnNode(node string, prefixes []string) error {
```
fedepaol, line 165: nit: let's call this `checkPrefixesOnNode`.

e2etests/tests/graceful_restart.go
```diff
@@ -133,3 +157,27 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
...
+	exc := executor.ForContainer(node)
```
fedepaol, line 166: sorry I just noticed this. If we do this, this test won't be able to run if/when we run on a real cluster. I think what other tests are doing (or, I hope) is to check on the frr-k8s pods. Should be equivalent as they are host - routed.

e2etests/tests/graceful_restart.go
```diff
@@ -68,19 +71,35 @@ var _ = ginkgo.Describe("Establish BGP session with EnableGracefulRestart", func
...
+
+		err := cleanup(updater)
+		Expect(err).NotTo(HaveOccurred(), "cleanup config in API and infra containers")
 	})
 
 	ginkgo.Context("When restarting the frrk8s deamon pods", func() {
-
-		ginkgo.DescribeTable("external BGP peer maintains routes", func(ipFam ipfamily.Family, prefix string) {
+		ginkgo.DescribeTable("both routes on nodes and on the external peers are maintained", func(ipFam ipfamily.Family, prefix string) {
+			learnPrefixes := []string{}
```
fedepaol, line 81: let's call this `advertisedPrefixes` or `externalPrefixes`

e2etests/tests/graceful_restart.go
karampok: good catch, but frr-k8s pods are restarting, so that is not an option.
In real cluster that would have to be ssh or find another hsot-routed pod (or create one)

fedepaol: Fair enough, no need to happen here.

fedepaol:
LGTM

PR #212, opened 2024-10-22
Support unnumbered BGP peering

karampok:
/kind feature

**What this PR does / why we need it**:
As the result of this design proposal
https://github.com/metallb/metallb/blob/main/design/unnumbered-bgp.md

**Special notes for your reviewer**:
Currently depends on this PR that adds the test functions needed
https://github.com/metallb/metallb/pull/2595,

**Release note**:

```release-note
Support unnumbered BGP peering
```

PR #239, opened 2024-12-12
E2E: bump e2etest version and use new field

karampok:
With the new version we need to change the .IP of the neighbor to .ID. The struct in metallb/e2etest looks as follows:

```
type Neighbor struct {
	// ID is the key that vtysh returns for the neighbor,
	// it can be either IP or interface name if unnumbered.
	ID                      string
	VRF                     string
```

 /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
NONE
```


karampok:
@oribon  this is needed because of the change we did on the other repo about `.IP` to become `.ID` and string.
Thanks!

karampok:
Given there are some other changes in the go.mod,
I bump as follow
```
#fix the go.mod to look to main
        go.universe.tf/e2etest => github.com/metallb/metallb/e2etest  main
```
and execute `go mod tidy` which fixes the version but as side effect some other stuff are needed.

karampok:
As @oribon suggested that to be merged either with the #212 or with #231

PR #270, opened 2025-03-13
Validate neighbors in same router,  add interface checks

karampok:
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```

### DIFF
--- a/internal/controller/api_to_config.go
+++ b/internal/controller/api_to_config.go
@@ -98,6 +98,10 @@ func apiToFRR(resources ClusterResources, alwaysBlock []net.IPNet) (*frr.Config,
 				return nil, err
 			}
 
+			if err := validateRouterConfig(routerCfg); err != nil {
+				return nil, err
+			}
+
 			curr, ok := routersForVRF[r.VRF]
 			if !ok {
 				routersForVRF[r.VRF] = routerCfg
@@ -706,4 +710,10 @@ func importedPrefixes(r v1beta1.Router, prefixesInRouter map[string][]string) ([
 		res = append(res, imported...)
 	}
 	return res, nil
+}
+
+func validateRouterConfig(r *frr.RouterConfig) error {
+	// merging with itself to validate neighbor list
+	_, err := mergeRouterConfigs(r, r)
+	return err
 }
--- a/internal/controller/api_to_config_test.go
+++ b/internal/controller/api_to_config_test.go
@@ -80,7 +80,6 @@ func TestConversion(t *testing.T) {
 								IPFamily:        ipfamily.IPv4,
 								Name:            "65002@192.0.2.2",
 								ASN:             "65002",
-								Port:            ptr.To[uint16](179),
 								SrcAddr:         "192.1.1.1",
 								Addr:            "192.0.2.2",
 								KeepaliveTime:   ptr.To[int64](20),
@@ -109,7 +108,7 @@ func TestConversion(t *testing.T) {
 									Neighbors: []v1beta1.Neighbor{
 										{
 											ASN:     65002,
-											Port:    ptr.To[uint16](179),
+											Port:    ptr.To[uint16](178),
 											Address: "192.0.2.2",
 											KeepaliveTime: &metav1.Duration{
 												Duration: 20 * time.Second,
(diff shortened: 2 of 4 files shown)


internal/controller/api_to_config.go
```diff
@@ -684,6 +688,20 @@ func validatePrefixes(prefixes []string) error {
 	return nil
 }
 
+func validateNeighbors(a []v1beta1.Neighbor) error {
```
oribon, line 691: this seems to cover only the specific case of mismatch between address and asns, while iirc there are more fields that can cause a similar conflict. would validating by merging the router with itself achieve the same goal?

internal/controller/merge.go
```diff
@@ -213,12 +213,16 @@ func routersAreCompatible(r, toMerge *frr.RouterConfig) error {
 	return nil
 }
 
-// Verifies that two neighbors are compatible for merging, assuming they belong to the same router.
```
oribon, line 216: why did you remove the `assuming..` part?

internal/controller/merge.go
```diff
@@ -213,12 +213,16 @@ func routersAreCompatible(r, toMerge *frr.RouterConfig) error {
...
+// Verifies that two neighbors are compatible for merging.
 func neighborsAreCompatible(n1, n2 *frr.NeighborConfig) error {
-	if n1.Addr != n2.Addr {
+	if n1.Addr != "" && n1.Addr != n2.Addr {
```
oribon, line 218: can this be `&& n2.Addr != "" && n1.Addr != n2.Addr` instead?

internal/controller/merge.go
```diff
@@ -213,12 +213,16 @@ func routersAreCompatible(r, toMerge *frr.RouterConfig) error {
...
 		return fmt.Errorf("neighbors with different addresses (%s != %s) are not compatible for merging", n1.Addr, n2.Addr)
 	}
 
+	if n1.Iface != "" && n1.Iface == n2.Iface {
```
oribon, line 222: same (with the &&s), also why isn't this `iface != iface`? we merge when the neighbor is the same

internal/controller/merge.go
karampok: I think this comment does not add value, it works toward confusing me, but no strong opinion I can add back

internal/controller/api_to_config.go
karampok: Refactored to re-use the merging logic

internal/controller/merge.go
oribon: I prefer we add it back, because the function really does make sense only when the neighbors belong to the same BGP router 😅

internal/controller/merge.go
```diff
@@ -213,17 +213,13 @@ func routersAreCompatible(r, toMerge *frr.RouterConfig) error {
 	return nil
 }
 
-// Verifies that two neighbors are compatible for merging, assuming they belong to the same router.
+// neighborsAreCompatible verifies that two neighbors are compatible for merging.
 func neighborsAreCompatible(n1, n2 *frr.NeighborConfig) error {
-	if n1.Addr != n2.Addr {
-		return fmt.Errorf("neighbors with different addresses (%s != %s) are not compatible for merging", n1.Addr, n2.Addr)
+	if n1.ID() != n2.ID() {
+		return fmt.Errorf("bug only neighbors with same ID are compatible for merging")
```
oribon, line 219: let's leave a `we shouldn't reach this` comment on the `if` instead of the "bug", and format the error with the different IDs similar to before.

internal/controller/api_to_config.go
```diff
@@ -98,6 +98,11 @@ func apiToFRR(resources ClusterResources, alwaysBlock []net.IPNet) (*frr.Config,
 				return nil, err
 			}
 
+			// merging with itself to validate neighbor list
+			if _, err := mergeNeighbors(routerCfg.Neighbors, routerCfg.Neighbors); err != nil {
+				return nil, err
+			}
```
oribon, line 104: let's wrap this in a `validateRouter` func, and also can we use `mergeRouterConfigs` instead?

internal/controller/api_to_config_test.go
```diff
@@ -2807,6 +2806,87 @@ func TestConversion(t *testing.T) {
 			secrets: map[string]v1.Secret{},
 			err:     errors.New("a not nil error"),
 		},
+		{
+			name: "Single router two neighbors different address different ASN",
```
oribon, line 2810: this is probably redundant

karampok: removed

internal/controller/api_to_config.go
karampok: done

internal/controller/merge.go
karampok: done

oribon (commented):
lgtm

PR #282, opened 2025-03-28
E2E: do not assert on the R-bit for sessions with Graceful Restart

karampok:
The R-bit might false depend on the FRR version and we do not control the value. So we avoid that check as it does not really add value.

From https://datatracker.ietf.org/doc/html/rfc4724

Restart Flags:

  This field contains bit flags related to restart.

      0 1 2 3
    +-+-+-+-+
    |R|Resv.|
    +-+-+-+-+

  The most significant bit is defined as the Restart State (R)
  bit, which can be used to avoid possible deadlock caused by
  waiting for the End-of-RIB marker when multiple BGP speakers
  peering with each other restart.  When set (value 1), this bit
  indicates that the BGP speaker has restarted, and its peer MUST
  NOT wait for the End-of-RIB marker from the speaker before
  advertising routing information to the speaker.

  The remaining bits are reserved and MUST be set to zero by the
  sender and ignored by the receiver.

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
 /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/tests/session.go
+++ b/e2etests/tests/session.go
@@ -257,7 +257,6 @@ var _ = ginkgo.Describe("Session", func() {
 					Expect(n.GRInfo.LocalGrMode).Should(Equal("Restart"))
 					Expect(n.GRInfo.RemoteGrMode).Should(Equal("Helper"))
 					Expect(n.GRInfo.NBit).To(BeTrue())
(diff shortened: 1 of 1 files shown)


oribon:
lgtm

PR #284, opened 2025-03-28
Webhook pod: report ready / live when the webhook endpoint is ready

fedepaol:

 /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Instead of relying on the metrics port, we move to test the webhook endpoint. The webhook will declare itself ready only when the cert controller did its part and the certificate is generated correctly.

This allows the crs not to fail on creation if we apply them after checking all pods are ready (ie right after a make deploy).

Also, make the webhook port configurable.

**Release note**:

```release-note
Declare the webhook pod to be ready only when it can serve requests, checking the webhook endpoint as readiness probe.
```


oribon (commented):
lgtm

Issue #285, opened 2025-03-31
BGPSessionState E2E flaking


oribon:
sometimes the test fail with:
```
"expected bfd status for node frr-k8s-control-plane peer 172.18.0.6 vrf  to be N/A, got Down"
```

e.g https://github.com/metallb/frr-k8s/actions/runs/14129016985/job/39584792360?pr=282

happens because of: https://github.com/FRRouting/frr/issues/18357

will be fixed when we move to FRR 10.x / we implement in the test's BeforeEach a validation that the previous config was actually deleted

PR #286, opened 2025-04-02
E2E: Fix BGPSessionState CI flake

oribon:
There is currently a bug on FRR 9.x that BFD profiles are not removed properly when a config that has BFD is replaced by one that does not, causing these tests to fail when the Clean isn't applied before the
tests themselves (because of the debouncer).
We put a temporary fix for that, to be removed
once we move to FRR 10.x or the bug is backported to FRR 9.

Fixes https://github.com/metallb/frr-k8s/issues/285

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
/kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/tests/bgp_session_status.go
+++ b/e2etests/tests/bgp_session_status.go
@@ -73,6 +73,7 @@ var _ = ginkgo.Describe("BGPSessionState", func() {
 		}
 		err = updater.Clean()
 		Expect(err).NotTo(HaveOccurred())
+		waitForFRRCleanState(cl) // TODO: remove when we move to FRR 10.x
 
 		cs = k8sclient.New()
 	})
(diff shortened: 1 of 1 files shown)


karampok:
Wondering if we could have the (BFD) tests cleanup their resources, or ` when a config that has BFD is replaced by one that does not` is actually the a test case?

e2etests/tests/bgp_session_status.go
```diff
@@ -408,3 +409,21 @@ func statusFormatFor(ip string) string {
 func ipFor(statusIP string) string {
 	return strings.ReplaceAll(statusIP, "-", ":")
 }
+
+// Ensures that all the frr-k8s daemons run a clean FRR config.
+// TODO: this is needed for a CI flake, remove once we move to FRR 10.x
+func ensureCleaned(cl client.Client) {
```
karampok, line 415: I would name it  something `WaitFRRNodeStatusesToBeCleaned` just not generic `ensureCleaned` that will force the reader to jump there

e2etests/tests/bgp_session_status.go
```diff
@@ -408,3 +409,21 @@ func statusFormatFor(ip string) string {
...
+	Eventually(func() error {
```
karampok, line 416: Ideally you could add `  GinkgoHelper()` in the  function

From https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure

Any function in which GinkgoHelper() is called is tracked by Ginkgo and ignored when a failure location is being computed. This allows you to build reusable test helpers and trust that the location presented to the user will always be in the spec that called the helper, and not the helper itself.

oribon: done (although I'm not familiar with how this affects the output)

e2etests/tests/bgp_session_status.go
oribon: done

oribon:
> Wondering if we could have the (BFD) tests cleanup their resources, or ` when a config that has BFD is replaced by one that does not` is actually the a test case?

It's not about the cleanup of the resources, because it was already there in the BeforeEach. The (FRR) bug is about one config containing BFD switching to another without, and handling this is the most bulletproof way to ensure these tests run in a clean state.

e2etests/tests/bgp_session_status.go
karampok: If fails, ginkgo will tell you it failed line 76, and not line 417

karampok (approved):

PR #289, opened 2025-04-10
Helm: add nodeSelector option for webhook

larainema:

>
> /kind bug
> /kind cleanup

/kind feature

> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Differentiate node selection for webhook server and frr-k8s DaemonSet. Webhook server now runs exclusively on control plane nodes.

**Release note**:

```release-note

```

### DIFF
--- a/charts/frr-k8s/README.md
+++ b/charts/frr-k8s/README.md
@@ -71,6 +71,7 @@ Kubernetes: `>= 1.19.0-0`
 | frrk8s.tolerateMaster | bool | `true` |  |
 | frrk8s.tolerations | list | `[]` |  |
 | frrk8s.updateStrategy.type | string | `"RollingUpdate"` |  |
+| frrk8s.webhook.nodeSelector | object | `{}` |  |
 | frrk8s.webhookPort | int | `9443` |  |
 | fullnameOverride | string | `""` |  |
 | nameOverride | string | `""` |  |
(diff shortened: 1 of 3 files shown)


fedepaol:
Hey! Sorry for the delay in reviewing this.
If I am not forgetting something, the webhook scheduling follows the same rules as frr-k8s: https://github.com/metallb/frr-k8s/blob/main/charts/frr-k8s/templates/webhooks.yaml#L100
You can schedule it on master too or let it run on any node, but we are not telling it anywhere to run ONLY on control plane nodes.

Is there a use case for needing to run the webhook pod on a particular set of nodes? Or your case is only to skip masters? If it's the latter, I'd change the logic to something like "runWebhookControllerOnMasters", which is similar to what we have now but for all the pods.

larainema:
> Hey! Sorry for the delay in reviewing this. If I am not forgetting something, the webhook scheduling follows the same rules as frr-k8s: https://github.com/metallb/frr-k8s/blob/main/charts/frr-k8s/templates/webhooks.yaml#L100 You can schedule it on master too or let it run on any node, but we are not telling it anywhere to run ONLY on control plane nodes.
> 
> Is there a use case for needing to run the webhook pod on a particular set of nodes? Or your case is only to skip masters? If it's the latter, I'd change the logic to something like "runWebhookControllerOnMasters", which is similar to what we have now but for all the pods.

our use case here is just let the webhook pod run only on control plane nodes, and we run frr-k8s strictly on the nodes that have the openvswitch label.

PR #290, opened 2025-04-14
Consume the FRRConfiguration from our namespace only

fedepaol:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

We should consume only the frrconfiguration instances belonging to our namespace and not from all the namespaces.

**Release note**:

```release-note
Ignore FRRConfigurations belonging to different namespaces.
```


oribon (commented):
in MetalLB we block creation via the webhook, is there a reason not to do it here also?

fedepaol:
> in MetalLB we block creation via the webhook, is there a reason not to do it here also?



> in MetalLB we block creation via the webhook, is there a reason not to do it here also?

I'd say this is more reliable. Won't hurt adding the logic in hte webhook too, so the user will understand why his config is not being processed.

fedepaol:
> > in MetalLB we block creation via the webhook, is there a reason not to do it here also?
> 
> > in MetalLB we block creation via the webhook, is there a reason not to do it here also?
> 
> I'd say this is more reliable. Won't hurt adding the logic in hte webhook too, so the user will understand why his config is not being processed.

Also, this is upgrade - proof.

fedepaol:
.. and now, if I add the webhook we won't be able to test the upgrade scenario :sweat_smile: 

either I leave the test as it is, providing coverage, or I enable the webhook and we trust upgrade works (I don't think adding an upgrade test is worth the effort)

oribon (commented):
lgtm

PR #296, opened 2025-04-28
Remove the disableMP behavior

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

As described in https://github.com/metallb/metallb/issues/2704, the genesis of the disableMP flag is a bit unfortunate.

"When moving to FRR, we inadvertently enabled the possibility to advertise the v6 addresses over a v4 session (and viceversa). This was totally unintentional, we tried to amend and then we found out that users (at least one!) were already relying on it.

This is causing confusion and works in a limited set of scenarios, because FRR doesn't have a guarantee of what the v6 nexthop should be for a session established over v6."

Because of this, we rollback the behavior, enabling an ipfamily to a neighbor connected with the corresponding ip family.

We also add a flag "dualstackipfamily" to enable the behavior for those scenarios where a user might want to enable a dual stack family for a neighbor.

**Release note**:

```release-note
Deprecate the disableMP flag, and make the v4 over v4 and v6 over v6 behavior the default. Additionally, add a dualStackIPFamily flag per neighbor, where we enable the v4 family over a v6 link and viceversa.
```


api/v1beta1/frrconfiguration_types.go
```diff
@@ -190,9 +192,16 @@ type Neighbor struct {
...
+	// Deprecated: DisableMP is deprecated in favor of DualStackIPFamily.
 	// +optional
 	// +kubebuilder:default:=false
 	DisableMP bool `json:"disableMP,omitempty"`
+
+	// To set if we want to enable the neighbor not only for the ipfamily related to its session,
+	// but also the other one. This allows to advertise/receive IPv4 prefixes over IPv6 sessions and vice versa.
+	// +optional
+	// +kubebuilder:default:=false
+	DualStackIPFamily bool `json:"dualStackIPFamily,omitempty"`
```
oribon, line 204: `DualStackIPFamily` seems a bit vague imo, especially when IPFamily shows up a lot around services 😅  is there something more specific we can have, maybe around address families? in any case I think having this as Family and an enum (SingleStack/DualStack) better than booleans.
also, is there value in making this more configurable? such as (possibly enabling) having a finer control over advertise/receive?

api/v1beta1/frrconfiguration_types.go
```diff
@@ -190,9 +192,16 @@ type Neighbor struct {
 	ToReceive Receive `json:"toReceive,omitempty"`
 
 	// To set if we want to disable MP BGP that will separate IPv4 and IPv6 route exchanges into distinct BGP sessions.
+	// Deprecated: DisableMP is deprecated in favor of DualStackIPFamily.
 	// +optional
 	// +kubebuilder:default:=false
 	DisableMP bool `json:"disableMP,omitempty"`
```
oribon, line 198: should we warn in the webhook when it's specified?

internal/controller/merge.go
```diff
@@ -245,8 +245,8 @@ func neighborsAreCompatible(n1, n2 *frr.NeighborConfig) error {
 		return fmt.Errorf("conflicting ebgp-multihop specified for %s", neighborKey)
 	}
 
-	if n1.DisableMP != n2.DisableMP {
-		return fmt.Errorf("conflicting disableMP specified for %s", neighborKey)
+	if n1.IPFamily != n2.IPFamily {
+		return fmt.Errorf("conflicting advertiseDualStack specified for %s", neighborKey)
```
oribon, line 249: nit: advertiseDualStack is probably a leftover

api/v1beta1/frrconfiguration_types.go
fedepaol: > `DualStackIPFamily` seems a bit vague imo, especially when IPFamily shows up a lot around services 😅 is there something more specific we can have, maybe around address families? in any case I think having this as Family and an enum (SingleStack/DualStack) better than booleans. 

Yeah but here it's frr-k8s, which has nothing to do with services. I can IPFamily + Enum, but I am not sure I can find a better name. Family is too vague imo.

> also, is there value in making this more configurable? such as (possibly enabling) having a finer control over advertise/receive?

Nope, because what we effectively do here is to enable / disable the neighbor per ip family (by adding the statement under the corresponding ipv4 / ipv6 family). That's also the reason why I went for a property of the neighbor vs a property of the advertisement. We could _theoretically_ add more complexity via route maps but I don't think it's worth it.

api/v1beta1/frrconfiguration_types.go
fedepaol: won't hurt

api/v1beta1/frrconfiguration_types.go
fedepaol: Also also, IPFamily is misleading as it should refer to ipv4, ipv6 OR dual stack.
I'll rename this to "dualstacksessoin" which is probably what makes sense in the frr context, and keep it as a bool, having a "sessionstack" still sounds clunky compared to force the sessoin to be dual stack

api/v1beta1/frrconfiguration_types.go
```diff
@@ -190,9 +192,16 @@ type Neighbor struct {
 	ToReceive Receive `json:"toReceive,omitempty"`
 
 	// To set if we want to disable MP BGP that will separate IPv4 and IPv6 route exchanges into distinct BGP sessions.
+	// Deprecated: DisableMP is deprecated in favor of DualStackIPFamily.
```
oribon, line 195: should be in favor of dualStackSession

api/v1beta1/frrconfiguration_types.go
oribon: now I'm a bit torn about using "session", because it could be misleading that the session itself is established over dual-stack (although there is only one neighborIP) 😅 I guess this is fine if you're okay with that and the alternatives seem clunkier

oribon (approved):
lgtm

PR #303, opened 2025-05-16
Deprecate disableMP - part 2

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Even though we were not enabling a given neighbor for the address family it didn't belong to, we were still defining all the prefix lists associated to it.

Here we reorganize the way the configuration is handled, by:
- moving away the logic from the template, which is difficult to follow compared to Go
- generating per address family prefix list and only after, it's related route map
- filtering prefixes if they don't match the address family of the neighbor 

**Release note**:

```release-note
FRR Configuration: don't generate the prefix lists if the neighbor is not enabled for that ip family. 
```


fedepaol:
cc @oribon , this is good to review

internal/controller/api_to_config.go
oribon, line 1: can we have this commit (3rd) together with the 2nd? it's easier to follow that way imo

internal/controller/api_to_config.go
```diff
@@ -255,111 +255,169 @@ func passwordForNeighbor(n v1beta1.Neighbor, passwordSecrets map[string]corev1.S
...
 
-// prefixesToMap returns two maps of prefix->OutgoingFilter (ie family, advertisement, communities), one for each family.
-// The ipv4Prefixes and ipv6Prefixes represent the "global" allowed prefixes which are the prefixes defined on the router.
-func prefixesToMap(toAdvertise v1beta1.Advertise, ipv4Prefixes, ipv6Prefixes []string) (map[string]*frr.OutgoingFilter, map[string]*frr.OutgoingFilter) {
-	resV4 := map[string]*frr.OutgoingFilter{}
-	resV6 := map[string]*frr.OutgoingFilter{}
-	if toAdvertise.Allowed.Mode == v1beta1.AllowAll {
-		for _, p := range ipv4Prefixes {
-			resV4[p] = &frr.OutgoingFilter{Prefix: p, IPFamily: ipfamily.IPv4}
+func prefixesWithLocalPrefToFRR(toAdd map[string]frr.LocalPrefPrefixList, neighbor *frr.NeighborConfig, toAdvertise v1beta1.Advertise, ipFamily ipfamily.Family, routerPrefixes sets.Set[string]) (map[string]frr.LocalPrefPrefixList, error) {
```
oribon, line 297: I think we can remove the toAdd here (or at least avoid re-using it inside the function)

internal/controller/api_to_config.go
```diff
@@ -255,111 +255,169 @@ func passwordForNeighbor(n v1beta1.Neighbor, passwordSecrets map[string]corev1.S
...
+
+		ipfamilyPrefixes := ipfamily.FilterPrefixes(prefixes.Prefixes, ipFamily)
+		for _, prefix := range ipfamilyPrefixes {
+			if !routerPrefixes.Has(prefix) {
+				return nil, fmt.Errorf("localPref %d associated to non existing prefix %s", prefixes.LocalPref, prefix)
+			}
+			if localPrefPrefixList.Prefixes.Has(prefix) {
+				return nil, fmt.Errorf("prefix %s is already defined for local preference %d", prefix, prefixes.LocalPref)
+			}
+			if existing, ok := toAdd[prefix]; ok && existing.LocalPref != prefixes.LocalPref {
```
oribon, line 323: I might be confused, but I think `toAdd[prefix]` will never be hit? seems like only keys in a different format get in this map (it's very probable I'm misinterpreting, because I assume we already have tests for this)
maybe this is already caught by validateOutgoingPrefixes and can be removed here?

internal/controller/api_to_config.go
```diff
@@ -457,6 +515,24 @@ func validateOutgoingPrefixes(prefixesInRouter []string, routerConfig v1beta1.Ro
 				return fmt.Errorf("trying to advertise non configured prefix %s to neighbor %s, vrf %s", p, neighborName(n), routerConfig.VRF)
 			}
 		}
+		localPrefForPrefix := map[string]uint32{}
+		for _, prefixes := range n.ToAdvertise.PrefixesWithLocalPref {
+			for _, p := range prefixes.Prefixes {
+				if ipfamily.ForCIDRString(p) == ipfamily.Unknown {
+					return fmt.Errorf("unknown ipfamily for prefix %s associated to localpref %d", p, prefixes.LocalPref)
+				}
```
oribon, line 523: this can use validatePrefixes ?

internal/controller/api_to_config.go
```diff
@@ -255,111 +255,169 @@ func passwordForNeighbor(n v1beta1.Neighbor, passwordSecrets map[string]corev1.S
...
 	}
+	return fmt.Sprintf("%s-%s-%s-community-prefixes", neighborID, comm, ipFamily)
+}
 
-	return nil
+func communityPrefixListKey(comm community.BGPCommunity, ipFamily string) string {
+	return fmt.Sprintf("%s-%s", comm, ipFamily)
+}
+
+func localPrefPrefixListKey(localPref uint32, ipFamily string) string {
```
oribon, line 419: can we rename ipfamily to frrIPFamily (and in the other functions we assume it is frr string)?

internal/controller/merge_test.go
```diff
@@ -1375,9 +1104,51 @@ func TestMergeNeighbors(t *testing.T) {
...
-			if diff := cmp.Diff(merged, test.expected, cmpopts.EquateEmpty()); diff != "" {
+			if diff := cmp.Diff(merged, test.expected,
+				cmpopts.EquateEmpty(), cmp.Comparer(communityComparer)); diff != "" {
 				t.Fatalf("config different from expected: %s", diff)
 			}
 		})
 	}
 }
+
+func communityPrefixListFor(neigID, comm string, ipFamily string, prefixes ...string) frr.CommunityPrefixList {
```
oribon, line 1115: can prefixes be []string? a bit easier to follow

internal/controller/merge_test.go
```diff
@@ -1375,9 +1104,51 @@ func TestMergeNeighbors(t *testing.T) {
...
+
+func keyForCommunityPL(comm string, ipFamily string) string {
+	community, err := community.New(comm)
+	if err != nil {
+		panic(err)
+	}
+	return communityPrefixListKey(community, ipFamily)
+}
+
+func localPrefPrefixListFor(neigID string, localPref int, ipFamily string, prefixes ...string) frr.LocalPrefPrefixList {
```
oribon, line 1138: same

internal/controller/api_to_config_test.go
```diff
@@ -585,26 +552,19 @@ func TestConversion(t *testing.T) {
...
-										},
-										{
-											IPFamily:  ipfamily.IPv4,
-											Prefix:    "192.0.6.0/24",
-											LocalPref: 100,
-										},
+									PrefixesV4: []string{"192.0.2.0/24", "192.0.4.0/24", "192.0.6.0/24"},
+									PrefixesV6: []string{},
+									CommunityPrefixesModifiers: map[string]frr.CommunityPrefixList{
+										keyForCommunityPL("10:100", "ip"):             communityPrefixListFor("192.0.2.21", "10:100", "ip", "192.0.2.0/24", "192.0.4.0/24"),
```
oribon, line 558: why here we expect the neighborID to be without the `asn@`?

internal/controller/api_to_config_test.go
```diff
@@ -3018,6 +3018,156 @@ func TestConversion(t *testing.T) {
...
+								{
+									ASN: 65040,
+									ID:  "192.0.2.20",
+									Neighbors: []v1beta1.Neighbor{
+										{
+											ASN:     65041,
+											Address: "192.0.2.21",
+											ToAdvertise: v1beta1.Advertise{
+												Allowed: v1beta1.AllowedOutPrefixes{
+													Prefixes: []string{"192.0.2.0/24", "2001:db8::/64"},
```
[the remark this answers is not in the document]
fedepaol: And allowedAll on a dual prefix is a bit of a different beast, because the prefixes there are not associated to the given neighbor. In this sense, I'd filter when allow all and reject if the user sets something specific in the neighbor (allow some , local pref or community)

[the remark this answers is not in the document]
oribon: sounds good

internal/controller/api_to_config.go
fedepaol: not sure I understand what you mean, the point here is to append items there

internal/controller/api_to_config.go
fedepaol: right, we are already validating it outside. Removing

PR #305, opened 2025-05-21
Make the webhook pod hostnetworked

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

If a user mis-applies a frrconfiguration, there's a risk he might lock himself out as the validating webhook will go through the CNI.

If we make the wh hostnetworked, we'll limit this issue as the apiserver will reach the endpoint via the host.

**Release note**:

```release-note
Make the webhook pod host networked so it will be possible to apply frr configurations even if some route is breaking the CNI connectivity.
```


karampok:
>If a user mis-applies a frrconfiguration, there's a risk he might lock himself out as the validating webhook will go through the CNI.

do you have any concret example what config (applied in the node by frrk8s) can do that?

fedepaol:
> > If a user mis-applies a frrconfiguration, there's a risk he might lock himself out as the validating webhook will go through the CNI.
> 
> do you have any concret example what config (applied in the node by frrk8s) can do that?

if you use bgp for pod to pod traffic, then the api server won't be able to reach the wh for validating any change of a broken configuration

karampok:
like we use frrk8s to build a overlay network (pod to pod traffic), and you break the overlay due to a bad frrconfig  and then  pod api on node X cannot speak to pod webhook on node Y (but it will still be able to speak if webhook is hostnetwork?)

fedepaol:
> like we use frrk8s to build a overlay network (pod to pod traffic), and you break the overlay due to a bad frrconfig and then pod api on node X cannot speak to pod webhook on node Y (but it will still be able to speak if webhook is hostnetwork?)

yes

karampok:
If the overlay is broken, a lot of stuff can fail (e.g. the process on the hostnetwork pod cannot talk to the API ).
Nmstate operator which have the same risk (breaking the connectivity) have a check (ping gateway) and if that check fails they role back.

fedepaol:
> If the overlay is broken, a lot of stuff can fail (e.g. the process on the hostnetwork pod cannot talk to the API ). 

The api is exposed via a service that is resolved to hostnetworked pods, so in theory it should be able to.
I agree that we could implement more complex solutions to this, but this also sounds like a good compromise in terms of added complexity / benefit.

charts/frr-k8s/README.md
```diff
@@ -93,6 +93,7 @@ Kubernetes: `>= 1.19.0-0`
 | prometheus.serviceMonitor.metricRelabelings | list | `[]` |  |
 | prometheus.serviceMonitor.relabelings | list | `[]` |  |
 | prometheus.serviceMonitor.tlsConfig.insecureSkipVerify | bool | `true` |  |
+| prometheus.whmetricsPort | int | `7672` |  |
```
karampok, line 96: Do we need this port because now we open a hostport so the other value is taken?

charts/frr-k8s/values.yaml
```diff
@@ -26,6 +26,9 @@ prometheus:
   # port frr-k8s will listen on for metrics
   metricsPort: 7572
 
+  # port the frr-k8s validating webhook will listen on for metrics
+  whmetricsPort: 7672
+
   # if set, enables rbac proxy on frr-k8s to expose
   # the metrics via tls.
   secureMetricsPort: 9140
```
karampok, line 34: the setup with rback proxy, is it tested? the rbac proxy will also be hostnetwork

charts/frr-k8s/README.md
fedepaol: yes

charts/frr-k8s/values.yaml
fedepaol: you can check how the CI is invoked -> deploy-helm which sets a bunch of helm parameters. Looking at the chart, you can see that it's installed.

fedepaol: or even better, the webhook doesn't come with the kube-rbac proxy

fedepaol: the webhook is not selected by the service of the service monitor, it's safe to just disable the port then.

cmd/main.go
oribon, line 1: I think 9443 is a bit too general to use as default on the host network, should we change it to something more exotic?

oribon (approved):
lgtm

PR #309, opened 2025-06-12
E2E: Add a simple session status test that does not rely on polling

oribon:
We add simple coverage for the session status resource that does not rely on the poll-interval, so that we can verify the feature on clusters where we don't have control over it.

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

We add simple coverage for the session status resource that does
not rely on the poll-interval, so that we can verify the feature
on clusters where we don't have control over it.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etests/tests/bgp_session_status.go
+++ b/e2etests/tests/bgp_session_status.go
@@ -101,59 +101,145 @@ var _ = ginkgo.Describe("BGPSessionState", func() {
 		}
 		type bfdState string
 		const (
-			noBFD   bfdState = "N/A"
-			bfdDown bfdState = "Down"
-			bfdUp   bfdState = "Up"
+			noBFD      bfdState = "N/A"
+			bfdDown    bfdState = "Down"
+			bfdUp      bfdState = "Up"
+			bfdUnknown bfdState = "Unknown"
 		)
 		bgpTryingToConnect := bgpPhase{sets.New("Idle", "Connect", "Active")}
 		bgpEstablished := bgpPhase{sets.New("Established")}
+
+		validateStatusForNeigh := func(neighborIP string, nodes []corev1.Node, vrf string, expectedBGP bgpPhase, expectedBFD bfdState) error {
+			for _, n := range nodes {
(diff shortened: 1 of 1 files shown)


e2etests/tests/bgp_session_status.go
```diff
@@ -101,59 +101,145 @@ var _ = ginkgo.Describe("BGPSessionState", func() {
...
-			validateNoStatusForAll := func(neighbors []frrk8sv1beta1.Neighbor, nodes []corev1.Node) error {
-				for _, n := range neighbors {
-					err := validateNoStatusForNeigh(n.Address, nodes, p.vrf)
-					if err != nil {
-						return err
-					}
+			return nil
+		}
+
+		validateNoStatusForAll := func(neighbors []frrk8sv1beta1.Neighbor, nodes []corev1.Node, vrf string) error {
```
karampok, line 151: Any specific reason why not defined outside like a normal function?

oribon: it is convenient to have it with a narrow scope, e.g to consume the `bgpPhase` structs that (currently) make sense to have in a narrow scope.

oribon:
@karampok `BGPSessionState.*Each.*node.*manages.*its.*statuses.*` is an option

oribon:
> > @karampok `BGPSessionState.*Each.*node.*manages.*its.*statuses.*` is an option
> 
> yes that would be an option you probably do not need any more .* , but in general having simple regex e.g. `--skip DUALSTACK` this look ok, having a very large set of patterns https://github.com/openshift/metallb/blob/e61d70062fce871a1f8a0eb39857a90dcd97ebfa/openshift-ci/run_frrk8s_e2e.sh#L12-L13 does not look perfect. My ask if we could some description that allow easier skipping

I guess ideally we can have something like that (or/and use ginkgo labels), but for our current usecase (and considering that skip is already in place) I think we should just keep it this way

PR #319, opened 2025-07-15
FRR: log stdout instead of a log file

oribon:
We remove the workaround which is no longer needed, with this we won't have the log file that grows infinitely without rotation.

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
frr container: no longer contains a log file, logs are sent directly to stdout. this enables the logs to be rotated on the node.
```

### DIFF
--- a/charts/frr-k8s/templates/controller.yaml
+++ b/charts/frr-k8s/templates/controller.yaml
@@ -99,7 +99,7 @@ data:
     frr defaults traditional
     hostname Router
     line vty
-    log file /etc/frr/frr.log informational
+    log stdout informational
 ---
 apiVersion: apps/v1
 kind: DaemonSet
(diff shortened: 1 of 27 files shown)


oribon:
operator lane fail expected, the other lane unnumbered flake

charts/frr-k8s/templates/controller.yaml
```diff
@@ -295,20 +295,10 @@ spec:
...
           - -c
-          - |
-            /sbin/tini -- /usr/lib/frr/docker-start &
-            attempts=0
-            until [[ -f /etc/frr/frr.log || $attempts -eq 60 ]]; do
-              sleep 1
-              attempts=$(( $attempts + 1 ))
-            done
-            tail -f /etc/frr/frr.log
+          - /sbin/tini -- /usr/lib/frr/docker-start
```
karampok, line 301: That script has also debug mode, can we enable it ? it it worth it (like is there anything important there)

oribon: I guess we should consider it, is there any docs for it?

karampok: I do not know, the code maybe https://github.com/FRRouting/frr/blob/master/docker/centos-8/docker-start

oribon: where did you see it has a debug mode?

karampok: I think I used it once,  I  do not remember details. Maybe this https://github.com/FRRouting/frr/blob/fd7c5adfcdae74d1c3e83567ee5e14d5e7d5b1a2/tools/frrcommon.sh.in#L57

but is only question for investigation, not strong opinion to follow it up or not.

Issue #274, opened 2025-03-19
Replace seds in Makefile with a proper kustomize patch


oribon:
We currently run some seds when running `deploy-controller`, mostly to have some configurations in place before CI runs. It would be nicer if there was some sort of kustomize patch to configure these parameters, as the current way is pretty hacky.
https://github.com/metallb/frr-k8s/pull/257#discussion_r1969700443

PR #323, opened 2025-07-17
Remove the frrnodestatus from a node if the pod that generated it is down

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

We want to be consistent with the FRRNodeStatus resource and remove it in case the pod that generated it is not running on a node anymore.

**Release note**:

```release-note
Keep the frrnodestate of a given node aligned with the lifecycle of the pod that generated it.
```


karampok:
@fedepaol  one initial question 
It would not be better to monitor the frr-k8s daemonset selectors and simply delete frrnodestates if found on a node where the daemonset is not running? Like we care more about the expected state and not just what the pod is momentarily doing.

I am thinking that,  if for a reason the pod is crashing/rebooting/oem killed, that will create many frrnodestates and will make the client to act  (not sure who is the client and what is doing though, but pretty sure the logs will go crazy). If we monitor the daemonset, the frrnodestate will only be removed in the real scenario we do not want pod there.

What do you think?

cmd/main.go
```diff
@@ -86,6 +87,7 @@ func main() {
 	flag.StringVar(&logLevel, "log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
 	flag.StringVar(&nodeName, "node-name", "", "The node this daemon is running on.")
 	flag.StringVar(&namespace, "namespace", "", "The namespace this daemon is deployed in")
+	flag.StringVar(&podName, "pod-name", "", "The pod name of this pod")
```
karampok: nit: probaly the pod name where the process is running

cmd/main.go
```diff
@@ -166,6 +168,18 @@ func main() {
 				setupLog.Error(err, "unable to create", "webhooks")
 				os.Exit(1)
 			}
+
+			setupLog.Info("Starting node status cleaner controller")
```
karampok: question: is status and state to be exactly the same?

cmd/main.go
```diff
@@ -86,6 +87,7 @@ func main() {
 	flag.StringVar(&logLevel, "log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
 	flag.StringVar(&nodeName, "node-name", "", "The node this daemon is running on.")
 	flag.StringVar(&namespace, "namespace", "", "The namespace this daemon is deployed in")
+	flag.StringVar(&podName, "pod-name", "", "The pod name of this pod")
 	flag.StringVar(&webhookMode, "webhook-mode", "disabled", "webhook mode: can be disabled or onlywebhook if we want the controller to act as webhook endpoint only")
```
karampok: I think that var (and maybe in general), there should be a rename because that pod is not just a webhook, now is a cleaner. But yes this would need general refactor

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,134 @@
+/*
+Copyright 2023.
```
karampok: nit: 2025?

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	client.Client
+	Scheme *runtime.Scheme
+	Logger log.Logger
+}
+
+// +kubebuilder:rbac:groups=frrk8s.metallb.io,resources=frrnodestates,verbs=get;list;watch;delete
+// +kubebuilder:rbac:groups="",resources=pods,verbs=get;list;watch
+
+func (r *NodeStateCleaner) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	level.Info(r.Logger).Log("controller", "NodeStateCleaner", "start reconcile", req.NamespacedName.String())
```
karampok: My gopls keeps suggesting
```
req.NamespacedName.String())
                                                                                          └──── default: could remove embedded field "NamespacedName" from selector
```

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// NodeStateCleaner reconciles Pod objects to clean up FRRNodeState resources.
+type NodeStateCleaner struct {
+	client.Client
+	Scheme *runtime.Scheme
+	Logger log.Logger
+}
+
+// +kubebuilder:rbac:groups=frrk8s.metallb.io,resources=frrnodestates,verbs=get;list;watch;delete
```
karampok, line 32: Is it best practice to have that per controller? we should have already the rbac for that

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,134 @@
...
+			level.Info(r.Logger).Log("controller", "NodeStateCleaner", "deleting FRRNodeState", "name", nodeState.Name, "pod", podName)
+			if err := r.Client.Delete(ctx, &nodeState); err != nil {
+				level.Error(r.Logger).Log("controller", "NodeStateCleaner", "failed to delete FRRNodeState", "name", nodeState.Name, "error", err)
+				return ctrl.Result{}, fmt.Errorf("failed to delete FRRNodeState: %w", err)
+			}
+			continue
+		}
+		if err != nil {
+			level.Error(r.Logger).Log("controller", "NodeStateCleaner", "failed to get pod", "pod", podName, "error", err)
+			return ctrl.Result{}, err
```
karampok: I think we should log and continue here,

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,134 @@
...
+		startupChan <- NewStartupEvent()
+	}()
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WatchesRawSource(source.Channel(startupChan, &handler.EnqueueRequestForObject{})).
+		WithEventFilter(predicate.Funcs{
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				// Only process frr-k8s pods
+				return e.Object.GetLabels()["app"] == "frr-k8s"
```
karampok: that might be a hardcoded dependency, is it possible in helm not to exist? https://github.com/metallb/frr-k8s/blob/f5b5b6796fd21f589fcc5d0b2b1f3f9e1bd275b2/charts/frr-k8s/templates/controller.yaml#L109-L114

config/testenv/default/frr-k8s_testenv_patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+        - "--namespace=$(NAMESPACE)"
+        - "--log-level=debug"
+        - "--pod-name=$(POD_NAME)"
+        - "--always-block=192.167.9.0/24,fc00:f553:ccd:e799::/64"
+      - name: frr-status
+        args:
+        - "--node-name=$(NODE_NAME)"
+        - "--namespace=$(NAMESPACE)"
+        - "--pod-name=$(POD_NAME)"
+        - "--poll-interval=5s"
```
karampok: space at the end

config/testenv/prometheus/frr-k8s_testenv_patch.yaml
```diff
@@ -0,0 +1,23 @@
...
+        - "--namespace=$(NAMESPACE)"
+        - "--log-level=debug"
+        - "--pod-name=$(POD_NAME)"
+        - "--always-block=192.167.9.0/24,fc00:f553:ccd:e799::/64"
+      - name: frr-status
+        args:
+        - "--node-name=$(NODE_NAME)"
+        - "--namespace=$(NAMESPACE)"
+        - "--pod-name=$(POD_NAME)"
+        - "--poll-interval=5s"
```
karampok: space at the end

config/testenv/default/kustomization.yaml
```diff
@@ -0,0 +1,14 @@
...
+resources:
+- ../../default
+
+patches:
+- path: frr-k8s_testenv_patch.yaml
+  target:
+    kind: DaemonSet
+    name: frr-k8s-daemon
+  options:
+    allowNameChange: false
```
karampok, line 12: empty lines bellow

internal/controller/frrstate_controller.go
```diff
@@ -137,6 +137,12 @@ func (r *FRRStateReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		}
 		return true
 	})
+
+	// Trigger initial reconcile event to create empty status
+	go func() {
+		r.Update <- NewStateEvent()
```
karampok, line 136: just a question: is it possible to  test that?

internal/controller/nodestate_cleaner_controller.go
fedepaol: lol AI :shrug:

internal/controller/frrstate_controller.go
fedepaol: yes, and there's a test doing that.

fedepaol:
> I would prefer to have single commit says add nodestatecleaner controller rather than three.

what are the three commits?
Asking because I am not sure. I can see the podname manifests one going together with the controller, not sure about the other one.

cmd/main.go
fedepaol: yeah, let me use state which matches the crd

cmd/main.go
fedepaol: what var?

internal/controller/nodestate_cleaner_controller.go
fedepaol: I think it's nice to know which rbac a controller needs, and we won't dupe them anyway

internal/controller/nodestate_cleaner_controller.go
fedepaol: that's debatable. If the get fails because of a transient error, it will be recovered as by returning error we exp backoff.
If we continue, we'd loose that pod until the next reconcile (for a new deletion).

internal/controller/nodestate_cleaner_controller.go
fedepaol: valid point, which made me notice we are giving cluster role permissions to pod.
Removed the selector, changed the pod permissions to be local to our ns

cmd/main.go
karampok: The var "webhookMode", now is more webhookAndCleanerMode, (as well as the pod/deployment etc)

internal/controller/frrstate_controller.go
karampok:
```
git show ce88bd9
commit ce88bd953e0a3a0b0e85a9260fca300f7ca66023
Author: Federico Paolinelli <fpaoline@redhat.com>
Date:   Thu Jul 17 11:26:26 2025 +0200

    Node state controller: generate a first empty status

    We want the status to say that there's a frrk8s instance available, so
    it makes sense to generate an empty one as soon as it starts.

    Signed-off-by: Federico Paolinelli <fpaoline@redhat.com>

diff --git a/internal/controller/frrstate_controller.go b/internal/controller/frrstate_controller.go
index 3c9cec3..138829a 100644
--- a/internal/controller/frrstate_controller.go
+++ b/internal/controller/frrstate_controller.go
@@ -137,6 +137,12 @@ func (r *FRRStateReconciler) SetupWithManager(mgr ctrl.Manager) error {
                }
                return true
        })
+
+       // Trigger initial reconcile event to create empty status
+       go func() {
+               r.Update <- NewStateEvent()
+       }()
+
```
I see that is was added in a single commit, and that commit did not have test. 


(small note: according to the AI feedback that implementation might have timing issues but tbo I cannot really see them)

fedepaol: yep, because this makes sense to be e2e tested more than unit tested. Hope the test is enough.
I can't think about timing issues here :sweat_smile:

fedepaol: Talking about here https://github.com/metallb/frr-k8s/pull/323/commits/aaeba53c5d616dda3366e258b6d29925e5e305f6#diff-98945e7f28b17e4edb69adfa63408c63af58e94de252dae83a846dc43bf6ec3cR427 where we verify a new node state is generated after zapping the pod.

internal/controller/nodestate_cleaner_controller.go
karampok: right, then I suppose we should log, store the error continue, and then when loop finishes return the stored error

cmd/main.go
fedepaol: I'll keep the pod name webhook, but will refine the parameter adding a new one specific to enable this controller, I think it's a good tradeoff

fedepaol:
ok, should've fixed all the comments and made some changes to fix CI failures.

charts/frr-k8s/templates/webhooks.yaml
```diff
@@ -36,6 +36,7 @@ spec:
         - --log-level={{ . }}
         {{- end }}
         - "--webhook-mode=onlywebhook"
+        - "--enable-node-state-cleaner=true"
```
karampok: Why optional? just for safety?

cmd/main.go
```diff
@@ -88,6 +91,7 @@ func main() {
 	flag.StringVar(&params.logLevel, "log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
 	flag.StringVar(&params.nodeName, "node-name", "", "The node this daemon is running on.")
 	flag.StringVar(&params.namespace, "namespace", "", "The namespace this daemon is deployed in")
+	flag.StringVar(&params.podName, "pod-name", "", "The name of the pod this process is running in")
```
karampok: Should we make this not optional? to fail if there is not pod-name argument? 
BUt I suppose the answer is that we do not do that for  neither nodeName or namespace .
Wondering what happens if someone updates the image without the manifest (probably nothing because cleaner ignore if no pod annotation)

cmd/main.go
```diff
@@ -132,7 +138,9 @@ func main() {
 		},
 		PprofBindAddress: params.pprofAddr,
 	}
+	enableFRRK8sControllers := params.webhookMode != "onlywebhook"
```
karampok: Can I  suggest a switch case which simplifies the general mode? like

```go
		switch params.mode {
		case "onlywebhook":
	    	   setupWebhook(mgr, logger)
		   startNodeStateCleaner(mgr, logger)
		case "nowebhook":
                   startFRRControllers(ctx, mgr, params, logger)
                 default
	    	   setupWebhook(mgr, logger)
		   startNodeStateCleaner(mgr, logger)
		   startFRRControllers(ctx, mgr, params, logger)
		}
```

I miss why in frr-k8s we do not have webhookMode boolean (where true means onlywebhook), when we want the controller to act as webhook, the default case in the above switch. Thanks

cmd/main.go
```diff
@@ -168,52 +176,12 @@ func main() {
...
 			}
-			return // We currently support only a onlywebhook mode
 		}
-
-		setupLog.Info("Starting controllers")
-		reloadStatusChan := make(chan event.GenericEvent)
-		reloadStatus := func() {
-			reloadStatusChan <- controller.NewStateEvent()
+		if params.enableNodeStateCleaner {
+			startNodeStateCleaner(mgr, logger)
```
karampok: here there are different approach: for webhook we log, check err, we exit. For the other two we do it internally, ideally should be the same design, no?
```
		if enableWebhook {
			setupLog.Info("Starting webhooks")
			err := setupWebhook(mgr, logger)
			if err != nil {
				setupLog.Error(err, "unable to create", "webhooks")
				os.Exit(1)
			}
		}
		if params.enableNodeStateCleaner {
			startNodeStateCleaner(mgr, logger)
		}
		if enableFRRK8sControllers {
			startFRRControllers(ctx, mgr, params, logger)
		}
```

config/rbac/nodestate_cleaner_role_binding.yaml
```diff
@@ -0,0 +1,20 @@
...
+  name: nodestate-cleaner-rolebinding
+  namespace: system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: nodestate-cleaner-role
+subjects:
+- kind: ServiceAccount
+  name: daemon
+  namespace: system
```
karampok: image
can  have this sign removed?

internal/controller/nodestate_cleaner_controller.go
karampok: Any argument against finishing the for loop and return the stored error instead of nil if there is an error? (not that we have high performance load operator)

cmd/main.go
```diff
@@ -66,49 +66,51 @@ func init() {
...
-	flag.StringVar(&namespace, "namespace", "", "The namespace this daemon is deployed in")
-	flag.StringVar(&webhookMode, "webhook-mode", "disabled", "webhook mode: can be disabled or onlywebhook if we want the controller to act as webhook endpoint only")
-	flag.BoolVar(&disableCertRotation, "disable-cert-rotation", false, "disable automatic generation and rotation of webhook TLS certificates/keys")
-	flag.BoolVar(&restartOnRotatorSecretRefresh, "restart-on-rotator-secret-refresh", false, "Restart the pod when the rotator refreshes its cert.")
-	flag.StringVar(&certDir, "cert-dir", "/tmp/k8s-webhook-server/serving-certs", "The directory where certs are stored")
-	flag.StringVar(&certServiceName, "cert-service-name", "frr-k8s-webhook-service", "The service name used to generate the TLS cert's hostname")
-	flag.StringVar(&pprofAddr, "pprof-bind-address", "", "The address the pprof endpoints bind to.")
-	flag.StringVar(&alwaysBlockCIDRs, "always-block", "", "a list of comma separated cidrs we need to always block")
-	flag.IntVar(&webhookPort, "webhook-port", 19443, "the port we listen the webhook calls on")
+	params := params{}
```
karampok: linter warning
```
params·:=·params{}
        └──── default: declaration of "params" shadows declaration at line 70
```
same name

cmd/main.go
```diff
@@ -166,62 +176,83 @@ func main() {
...
+	}
+	frrInstance := frr.NewFRR(ctx, reloadStatus, logger, logging.Level(params.logLevel))
+
+	alwaysBlock, err := parseCIDRs(params.alwaysBlockCIDRs)
+	if err != nil {
+		setupLog.Error(err, "failed to parse the always-block parameter", "always-block", params.alwaysBlockCIDRs)
+		os.Exit(1)
+	}
+
+	dumpResources := false
```
karampok: lint warning
```
󰌶   1 » dumpResources·:=·false
        └──── default: could merge conditional assignment into variable declaration
```
maybe 
```
	dumpResources := params.logLevel == logging.LevelDebug || params.logLevel == logging.LevelAll
```

internal/controller/frrstate_controller.go
```diff
@@ -108,6 +122,7 @@ func (r *FRRStateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	}
 
 	state.Status = newStatus
+
```
karampok: empty line

internal/controller/frrstate_controller.go
```diff
@@ -90,12 +91,25 @@ func (r *FRRStateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	err := r.Client.Get(ctx, types.NamespacedName{Name: r.NodeName}, state)
 	if k8serrors.IsNotFound(err) {
 		state.Name = r.NodeName
+
+		state.Annotations = map[string]string{
```
karampok: I think we should extend the unit test file  frrstate_controller_test.go to test if pod owner annotations is setup, if that is replace properly

karampok (commented):
```
git lg -5 |grep -E "state|status"
* 879ddf1 - (HEAD -> removestatus) E2E: add a test where we check if the node state is cleaned (2 hours ago) <Federico Paolinelli>
* 821ab4d - Node state controller: generate a first empty status (2 hours ago) <Federico Paolinelli>
* 1813619 - Add an optional frrnodestatus cleanup controller (2 hours ago) <Federico Paolinelli>
```
can se keep state (frnodestatus)

charts/frr-k8s/templates/webhooks.yaml
fedepaol: because I don't want to tie it to the webhook pod, I thikn this is cleaner than calling the webhook "cleaner"

cmd/main.go
fedepaol: the image without manifest case is never supported

cmd/main.go
fedepaol: > Can I suggest a switch case which simplifies the general mode? like
> 
> ```go
> 		switch params.mode {
> 		case "onlywebhook":
> 	    	   setupWebhook(mgr, logger)
> 		   startNodeStateCleaner(mgr, logger)
> 		case "nowebhook":
>                    startFRRControllers(ctx, mgr, params, logger)
>                  default
> 	    	   setupWebhook(mgr, logger)
> 		   startNodeStateCleaner(mgr, logger)
> 		   startFRRControllers(ctx, mgr, params, logger)
> 		}
> ```
> 

I'd rather keep the cleaner independent from the webhook flag

> I miss why in frr-k8s we do not have webhookMode boolean (where true means onlywebhook), when we want the controller to act as webhook, the default case in the above switch. Thanks

tbh, I don't remember and goes beyond the purpouse of this PR

cmd/main.go
fedepaol: yep, that makes sense, will change

cmd/main.go
fedepaol: what linter are you using / can we enforce it in CI?

internal/controller/frrstate_controller.go
fedepaol: not sure I understood, mind expanding?

cmd/main.go
fedepaol: doesn't seem more readeable

fedepaol: plus, it was already like this, I just moved the code

internal/controller/nodestate_cleaner_controller.go
fedepaol: nope, will do

cmd/main.go
fedepaol: and, do you think it's critical given we use only an instance of params here? It was kind of intentional, instead of having to figure out a paramsStruct or some other name. Initially, params was anonymous but then I had to pass it, hence the structure definition

karampok: good question I am running latest gopls with nothing disabled in my vim
https://git.sr.ht/~karampok/nvim-config/tree/dev/item/lsp/gopls.lua#L68

Always lost what golangci give versus what gopls, I have a todo to update linter, and hopefully we could catch those (or agree we do not want to catch)

internal/controller/frrstate_controller.go
karampok: I mean that the test file has not reference to annotation
of if I add this
```
        if state.Annotations[PodNameAnnotation] != r.PodName {
+               if true {
+                       panic("x")
+               }
```
I can run the unit-tests without panic

I think setting the owner should be unit tested also not just e2e test

karampok:
> Sorry, mind expanding "can se keep state (frnodestatus)" ?

I mean only use the word state (frrnodestate), empty node state, `git lg -5 |grep status` to return no result

cmd/main.go
fedepaol: for now, I'd take for granted what the CI linting gives
.
If we need to expand the rules, fine. But it's either you think the linter can be improved OR you need to vlaidate against what we have here.

cmd/main.go
karampok: I think we do not need three modes like we do in metallb and we should one day clean it up.
But not the purpose of this PR.

oribon: > tbh, I don't remember and goes beyond the purpouse of this PR

iirc the main reason was certs rotator not happy with multiple pods, where we don't have this problem with metallb's single replica controller.

imo it makes sense to just separate the webhook+new cleaner into a different binary (in this PR or another)

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,133 @@
...
+			errors = append(errors, fmt.Errorf("failed to get pod %s: %w", podName, err))
+		}
+	}
+
+	if len(errors) > 0 {
+		level.Error(r.Logger).Log("controller", "NodeStateCleaner", "reconcile finished with errors", "error_count", len(errors))
+		return ctrl.Result{}, fmt.Errorf("reconcile finished with %d errors", len(errors))
+	}
+
+	return ctrl.Result{}, nil
```
oribon: this can be replaced with `utilerrors.NewAggregate(errs)`

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,133 @@
...
+func (r *NodeStateCleaner) SetupWithManager(mgr ctrl.Manager) error {
+	startupChan := make(chan event.GenericEvent, 1)
+	// Trigger initial reconciliation, as pod might have
+	// been deleted while we are down.
+	go func() {
+		startupChan <- NewStartupEvent()
+	}()
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Pod{}).
```
oribon, line 79: wdyt about squashing with `handler.EnqueueRequestsFromMapFunc` this (as was done for the frrconfiguration_controller)?

oribon, line 79: it _might_ be a stretch, but we could also watch create/deletes of nodestates (that way we also keep the relation between pod<->state e.g if a user creates a state for no reason, and we can drop the startupChan which is a bit ugly)

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,133 @@
...
+				return false
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				return false
+			},
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return true
+			},
+		}).
+		Complete(r)
```
oribon, line 97: I think we can add a predicate to only care about pod deletions

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -0,0 +1,133 @@
...
+
+	frrNodeStates := &frrk8sv1beta1.FRRNodeStateList{}
+	if err := r.Client.List(ctx, frrNodeStates); err != nil {
+		level.Error(r.Logger).Log("controller", "NodeStateCleaner", "failed to list FRRNodeStates", err)
+		return ctrl.Result{}, err
+	}
+
+	var errors []error
+	for _, nodeState := range frrNodeStates.Items {
+		podName, exists := nodeState.Annotations[PodNameAnnotation]
```
oribon: we don't really need this, as we can match nodestate name to pod's node (fetching pods with `app.kubernetes.io/component: frr-k8s` label). unless you really like the explicitness of the relation between the two controllers, I think removing it and just matching states to nodes with pods is cleaner

Makefile
oribon, line 1: reminder to close https://github.com/metallb/frr-k8s/issues/274 after this 😛

e2etests/tests/status.go
```diff
@@ -349,6 +350,95 @@ var _ = ginkgo.Describe("Exposing FRR status", func() {
...
+
+			err = cs.CoreV1().Pods(k8s.FRRK8sNamespace).Delete(context.Background(), podToDelete.Name, metav1.DeleteOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func() bool {
+				_, err := cs.CoreV1().Pods(k8s.FRRK8sNamespace).Get(context.Background(), podToDelete.Name, metav1.GetOptions{})
+				return errors.IsNotFound(err)
+			}, time.Minute, time.Second).Should(BeTrue())
+
+			ginkgo.By("Waiting for FRRNodeState to be deleted")
```
oribon: I might be missing something, but this test is possibly a vw? (since we don't really change the daemonset's nodeSelector)
we're checking if a state was deleted by going over all of the states and checking that none contains the old pod name, but it could be just that the new pod overwrote the previous state, so this test would pass even without the cleaner controller running

internal/controller/nodestate_cleaner_controller.go
fedepaol: which is this one..?

cmd/main.go
fedepaol: Yes, the question here is why we have the 3 modes while we only support no webhook / standalone.
And yes, I'll make a new binary.

internal/controller/nodestate_cleaner_controller.go
fedepaol: I'd say we should see this being a bottleneck before doing it.
We only listens to frrnodestate creations and pod deletions, which have a predictable cardinality (vs frrconfiguration which can effectively explode)

internal/controller/nodestate_cleaner_controller.go
fedepaol: that was my first iteration, but I didn't like the idea of harvesting all the pods looking for the right one. Additionally, we use a different app annotation in charts vs manifests.
This is more efficient and straightforward imo.

e2etests/tests/status.go
fedepaol: yep, valid point, it might be the case. I was hesitant in scaling down the ds/ as this test might run also under the operator, which would reconcile the daemonset. Let me add a count of node states, this should ensure the test is valid (or tell us if it's vw-ing)

internal/controller/nodestate_cleaner_controller.go
oribon: we can add add a new label to both if that's our problem, and efficiency/straightforwardness is debatable 😅 (2 controllers "interacting" through annotations vs listing all pods with a label)

internal/controller/nodestate_cleaner_controller.go
oribon: I was thinking more about alignment of the code, but as you prefer

karampok:
@fede you need this commit for the metallb_e2e to pass https://github.com/metallb/frr-k8s/pull/324/commits/013cdd8c4b0264f4b805cc35b5185012c46eb0e6

e2etests/tests/status.go
oribon: yeah I figured, I guess a fair compromise can be manually creating a nodestate pointing to a non-existing pod / with a bad node name (depending on the other discussion), and seeing it get deleted

charts/frr-k8s/templates/webhooks.yaml
```diff
@@ -29,14 +29,12 @@ spec:
       {{- end }}
       containers:
       - command:
-        - /frr-k8s
+        - /statuscleaner
```
karampok: Having a file called webhook.yaml, which installs a deployment called satuscleaner which run a binary called statuscleaner which serves also a webhook service is not very clean.

IMO if we want clean split we should have three  files, one called webhook, one called statuscleaner, one controller,  and each have each deployment with the right name and each pod have  a binary with  right name.

If we  want to go for one binary (for webhook/cleaner part) we should pick a generic name that is not tight to single function. E.g. manager (expecting manager pod can run webhook controller, or statuscleaner or some new in the future).

cmd/frr-k8s-controller/main.go
```diff
@@ -0,0 +1,201 @@
+/*
+Copyright 2023.
```
karampok, line 2: nit: one of the new files should be 2025

config/frr-k8s/frr-k8s.yaml
```diff
@@ -343,3 +341,4 @@ spec:
       serviceAccountName: daemon
       terminationGracePeriodSeconds: 10
       hostNetwork: true
+
```
karampok: empty line

karampok (commented):
About the commits I would expect the commit `Add an optional frrnodestate cleanup controller ` to include the binary split. Like
```
commit 1: split into two binaries wehook/controller
commit 2: add frrnodestate controller to webhook binary
````

Also commits in this PR adds/removes implementation   e.g. `enableNodeStateCleaner` which should not exist at all in git history, no?

nit: to be consistent on the path, I think prefix frr-k8s can be removed

````
❯ tree cmd
cmd
├── frr-k8s-controller  # this could just controller
│   ├── cidrs_test.go
│   └── main.go
└── statuscleaner #this could be also frr-k8s-statuscleaner
    └── main.go
````

charts/frr-k8s/templates/webhooks.yaml
fedepaol: > Having a file called webhook.yaml, which installs a deployment called satuscleaner which run a binary called statuscleaner which serves also a webhook service is not very clean.
> 
> IMO if we want clean split we should have three files, one called webhook, one called statuscleaner, one controller, and each have each deployment with the right name and each pod have a binary with right name.
> 

The only reason why we have a separate wh, is to be able to deploy the validation logic upfront (for example in the metallb operator). Otherwise, the wh would have been embedded in frrk8s pods. This is to say that I'd avoid to add more than one extra pod.

> If we want to go for one binary (for webhook/cleaner part) we should pick a generic name that is not tight to single function. E.g. manager (expecting manager pod can run webhook controller, or statuscleaner or some new in the future).


Manager says nothing and is not a meaningful name. I can call it webhook-cleaner, but it becomes too complicated. Imo webhook is it's main role and we can document somewhere that the frr state cleaning part is handled by it as it runs in signle isntance (if not obvious by looking at the code), or have another pod but I don't want that.
Imo webhook is better than a neutral name, like controller, manager or master. It probably makes sense to change back the name of the binary to webhook so it's consistent. I was dragged away by the new feature.

The other option is to stick with cleaner, stating that that's the responsibility of that pod. The webhook functionality is generally ancillary to some pod (or all the pods, as it would've been if we hadn't the operator problem). For example, in MetalLB it's part of the controller pod.

karampok: > Otherwise, the wh would have been embedded in frrk8s pods.

that would have been so nice :)

> Manager says nothing and is not a meaningful name. I can call it webhook-cleaner, but it becomes too complicated. Imo webhook is it's main role and we can document somewhere that the frr state cleaning part is handled by it as it runs in signle isntance (if not obvious by looking at the code), or have another pod but I don't want that.
Imo webhook is better than a neutral name, like controller, manager or master. It probably makes sense to change back the name of the binary to webhook so it's consistent. I was dragged away by the new feature.

> The other option is to stick with cleaner, stating that that's the responsibility of that pod. The webhook functionality is generally ancillary to some pod (or all the pods, as it would've been if we hadn't the operator problem). For example, in MetalLB it's part of the controller pod.

I do not think there is clearly right/wrong approach.  My main point is to have at least concistency on the name of file,deployment,binary (and clearly document that webhook binary has cleaner controller or the vice versa)

cmd/frr-k8s-controller/main.go
fedepaol: they both originate from a 2023 one, I'd keep it as it is

oribon:
lgtm

PR #325, opened 2025-07-21
CI: add a daily ci run

karampok:
Having a periodic run is needed because:
1. in our CI we fetch the metallb main branch, so if something break that will be only visible when a new PR will land in the repo.
2. flakes will be found.
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
NONE
```


karampok:
> This is true, but frr-k8s serves metallb, not the other way around. So, if a change in frr-k8s breaks metallb, we will found out in the PR itself. The other way around (metallb breaks frr-k8s) is not possible.

not sure I follow exactly. But is about the tooling around also,  for  example this commit https://github.com/metallb/frr-k8s/pull/325/commits/a9c3a7308806bb4a32805f7bdad45508122c4e74 broke the ci of this repo. The daily would have marked it red. 

> This is a bit of a weak argument, we are getting flakes already without the periodic.

Sure , we have flakes but if you have a flake that happens one out 30, that will be visible looking the nightlies (but will not be visible or ignored during your PR ci run) 

> Adding the periodics is not free, we need to maintain and observe them, and as of today I am not sure the benefit from that is worth it. Periodics make sense if there are external factors that can change, but since we pin basically everything, there's little value imo.

I think there is value, just by looking the the runs  in overall you can extract statistics. We could use AI to extract info withotu effort  (like which tests are talking long, which fail how often), as long we have the data. (also external factor is the ubuntu image that might be bumped, the image tags that might be replaced). Not sure what is the harm (as there is no cost AFAIK).

fedepaol:
> > This is true, but frr-k8s serves metallb, not the other way around. So, if a change in frr-k8s breaks metallb, we will found out in the PR itself. The other way around (metallb breaks frr-k8s) is not possible.
> 
> not sure I follow exactly. But is about the tooling around also, for example this commit [a9c3a73](https://github.com/metallb/frr-k8s/commit/a9c3a7308806bb4a32805f7bdad45508122c4e74) broke the ci of this repo. The daily would have marked it red.
> 

There is a substancial difference, which is:
 
we didn't break metallb, we "just" broke the frrk8s CI that runs metallb tests. 

Why we know we didn't break metallb? Because:
- when we file a PR against metallb, we run the metallb tests against a pinned frr-k8s
- when we file a PR against frr-k8s, we run metallb's main against the current frr-k8s

> > This is a bit of a weak argument, we are getting flakes already without the periodic.
> 
> Sure , we have flakes but if you have a flake that happens one out 30, that will be visible looking the nightlies (but will not be visible or ignored during your PR ci run)

Which, imo, makes it not worth pursuing.

> 
> > Adding the periodics is not free, we need to maintain and observe them, and as of today I am not sure the benefit from that is worth it. Periodics make sense if there are external factors that can change, but since we pin basically everything, there's little value imo.
> 
> I think there is value, just by looking the the runs in overall you can extract statistics. We could use AI to extract info withotu effort (like which tests are talking long, which fail how often), as long we have the data. (also external factor is the ubuntu image that might be bumped, the image tags that might be replaced). Not sure what is the harm (as there is no cost AFAIK).


I am not saying there is no value in this, I am saying the value is little (because we pin our deps) vs having to check yet another periodic (assuming that somebody must check the periodic, which I don't want to commit to).
 
The cost stays in following the reports, and having yet another thing to check.
The info you refer to can be extracted by the current history of jobs the same way.

PR #356, opened 2025-10-07
fix: duplicate port name "monitoring"

dave-tucker:

/kind cleanup

Deploying the all-in-one YAML results in the following warning.

Warning: spec.template.spec.containers[4].ports[0]: duplicate port name "monitoring" with spec.template.spec.containers[2].ports[0], services and probes that select ports by name will use spec.template.spec.containers[2].ports[0]

This hopefully fixes that by renaming the frr-metrics port from monitoring to frr-metrics.

**Release note**:

```release-note
Fix duplicate port name `monitoring` when deploying the all-in-one.yaml
```


dave-tucker:
@fedepaol sorted! thanks

PR #357, opened 2025-10-13
Release notes: automate the generation

fedepaol:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

The generation of the release notes was half baked, including a lot of manual steps.
Here we try to fully automate it by fetching the previous commit containing a release note and optinally committing it.

All the user needs to provide is the GITHUB_TOKEN env variable.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/gen_relnotes.sh
+++ b/hack/gen_relnotes.sh
@@ -1,5 +1,11 @@
 #!/bin/bash -e
-set -x
+
+# Script to automate generation and committing of release notes
+# Usage:
+#   ./hack/gen_relnotes.sh           # Automates entire release notes workflow (auto-increments version)
+#   ./hack/gen_relnotes.sh <version> # Automates workflow with custom version (e.g., 1.2.3)
+
+set -e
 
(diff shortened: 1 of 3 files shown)


hack/gen_relnotes.sh
```diff
@@ -1,32 +1,125 @@
 #!/bin/bash -e
-set -x
+
+# Script to automate generation and committing of release notes
+# Usage:
+#   ./hack/gen_relnotes.sh           # Automates entire release notes workflow (auto-increments version)
+#   ./hack/gen_relnotes.sh <version> # Automates workflow with custom version (e.g., 1.2.3)
+
+set -e
```
karampok, line 8: duplicate -e 
```
#! /bin/bash
set -euo pipefail
```
is safer to fail if env is not defined

fedepaol: will follow up in another pr

PR #369, opened 2025-11-13
Allow to use a different selector for the node state cleaner

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Depending on the deployment type, the frrk8s pods might have a different label, or the user might want to apply this logic only to some pods.

Here we make the label selector customizable.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/webhooks/frrconfiguration_webhook.go
+++ b/internal/webhooks/frrconfiguration_webhook.go
@@ -243,6 +243,7 @@ var getNodes = func() ([]corev1.Node, error) {
 func containsDisableMP(routers []v1beta1.Router) bool {
 	for _, r := range routers {
 		for _, n := range r.Neighbors {
+			//nolint:staticcheck // DisableMP is deprecated but still supported for backward compatibility
 			if n.DisableMP {
 				return true
 			}


cmd/statuscleaner/main.go
```diff
@@ -71,6 +71,7 @@ type params struct {
 	certServiceName               string
 	pprofAddr                     string
 	webhookPort                   int
+	FRRK8sSelector                string
```
karampok: This is the only one that starts with capital (exported), maybe we could have it small?

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -96,3 +98,19 @@ func (r *NodeStateCleaner) SetupWithManager(mgr ctrl.Manager) error {
 		}).
 		Complete(r)
 }
+
+type PodSelector struct {
```
karampok: Do we need really a new type? 
At the end can we have instead of 
```
	if err := r.List(ctx, pods, client.InNamespace(r.Namespace), client.MatchingLabels(map[string]string{
		r.FRRK8sSelector.Key: r.FRRK8sSelector.Value,
	})); err != nil {
```
something simpler like
```
	if err := r.List(ctx, pods, client.InNamespace(r.Namespace), frrk8sLabels); err != nil {
```
and avoid if possible custom parsing?

(a first (AI) search shows https://pkg.go.dev/k8s.io/apimachinery/pkg/labels this could help, but maybe are other solutions)

internal/controller/nodestate_cleaner_controller.go
```diff
@@ -96,3 +98,19 @@ func (r *NodeStateCleaner) SetupWithManager(mgr ctrl.Manager) error {
...
+	Key   string
+	Value string
+}
+
+func PodSelectorFromString(s string) (PodSelector, error) {
```
karampok: The string can finally be not valid k8s labels but that function will pass (and probably will just fail on the list).

https://stackoverflow.com/questions/75081380/generated-valid-label-value-kubernetes (random link)

cmd/statuscleaner/main.go
fedepaol: yep

internal/controller/nodestate_cleaner_controller.go
fedepaol: let me try to refactor this

internal/controller/nodestate_cleaner_controller.go
fedepaol: valid point, added

PR #372, opened 2025-12-10
Add basic EVPN support

jcaamano:
This PR aims to introduce basic EVPN configuration support to the FRRConfiguration CRD.

- [x] Design proposal
- [ ] API
- [ ] Webhook Validation
- [ ] Rendering
- [ ] Tests
- [ ] Docs 

/kind design

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/design/basic-evpn-support.md
+++ b/design/basic-evpn-support.md
@@ -1,0 +1,504 @@
+# EVPN Support for FRRConfiguration
+
+## Summary
+
+This proposal adds Ethernet VPN (EVPN) configuration support to the FRRConfiguration CRD in frr-k8s. EVPN is a BGP-based control plane for VXLAN overlays, enabling Layer 2 and Layer 3 VPN connectivity with better scalability and multi-tenancy support compared to traditional overlay protocols.
+
+The primary motivation is to support the [OVN-Kubernetes EVPN enhancement](https://github.com/ovn-kubernetes/ovn-kubernetes/pull/5089), which requires advertising User Defined Networks (UDNs) externally via EVPN. This enhancement enables OVN-Kubernetes to integrate with external networks using standard EVPN/VXLAN technology instead of proprietary overlay protocols.
+
+This proposal focuses on extending the FRRConfiguration API with EVPN-specific fields, adding validation, and implementing FRR configuration rendering. No architectural changes to frr-k8s are required.
+
+## Motivation
+
+OVN-Kubernetes is adding support for exposing User Defined Networks (UDNs) to external networks using EVPN. This enables:
+
+1. **External connectivity**: Kubernetes workloads can communicate with external networks (VMs, bare-metal, other clusters) using EVPN/VXLAN
+2. **Multi-tenancy**: Extending user defined networks isolation to the provider network allowing overlapping IP spaces without conflict
+
+Currently, frr-k8s only supports basic BGP unicast route advertisement. OVN-Kubernetes needs the ability to:
+
+- Configure BGP neighbors for the L2VPN EVPN address family
+- Advertise VXLAN Network Identifiers (VNIs) for both L2 and L3 VPNs
+- Configure route distinguishers (RD) for route traceability and route targets (RT) for controlling route distribution between VRFs
+- Advertise unicast prefixes as EVPN type-5 routes for L3 VPN connectivity
+
+### Goals
+
+- Add EVPN configuration to the FRRConfiguration CRD API
+- Support both Layer 2 VNI (MAC-VRF, bridging) and Layer 3 VNI (IP-VRF, routing) configuration
+- Enable BGP neighbor activation for the L2VPN EVPN address family
+- Support route distinguisher (RD) and route target (RT) configuration for VNIs
+
+### Non-Goals
+
+The following features are not required by OVN-Kubernetes and are therefore out of scope for this proposal:
+
+- **Route-map based filtering and customization**: Advanced route filtering and customization (prefix filtering, VNI filtering, route modification) via route-maps
+- **Advanced EVPN features**: Multi-homing, BUM handling customization, advertise-PIP, advertise-svi-ip, and other advanced EVPN capabilities beyond basic L2/L3 VPN functionality
+- **Alternative data planes**: This proposal only supports EVPN over VXLAN. Other data planes (MPLS, SR-MPLS, SRv6) are out of scope
+- **Host networking configuration**: Setting up the underlying host networking infrastructure (VXLAN interfaces, bridge/VRF configuration, VNI-to-interface mappings) is out of scope. External tooling or manual configuration will be required to configure the host networking as FRR expects for EVPN functionality
+
+## Proposal
+
+### User Stories
+
+As a **cluster administrator**, I want to:
+
+1. **Configure EVPN neighbors**: Enable BGP sessions with external EVPN routers (route reflectors or VTEPs) by activating the L2VPN EVPN address family on existing BGP neighbors.
+
+2. **Exchange L2 VNI routes**: Advertise and receive EVPN type-2 (MAC/IP Advertisement) and type-3 (Inclusive Multicast Ethernet Tag) routes for L2 VNIs, enabling Layer 2 connectivity to external networks.
+
+3. **Exchange L3 VNI routes**: Advertise and receive EVPN type-5 (IP Prefix) routes for L3 VNIs with explicitly configured unicast prefixes, enabling Layer 3 connectivity to external networks.
+
+4. **Customize route parameters**: Override the route distinguisher and route targets for VNIs to control route distribution and filtering.
+
+### Design Details
+
+#### API Changes
+
+The proposal adds an `evpn` field to the existing `Router` configuration and an `addressFamilies` field to the `Neighbor` configuration.
+
+**Router Configuration:**
+
+```yaml
+apiVersion: frrk8s.metallb.io/v1beta1
+kind: FRRConfiguration
+metadata:
+  name: evpn-example
+spec:
+  bgp:
+    routers:
+    - asn: 64512
+      neighbors:
+      # EVPN neighbor for L2VPN routes
+      - address: 192.168.1.1
+        asn: 64512
+        addressFamilies: ["evpn"]
+      # Unicast neighbor for standard routes
+      - address: 192.168.1.2
+        asn: 64512
+        addressFamilies: ["unicast"]
+        toAdvertise:
+          allowed:
+            mode: all
+        toReceive:
+          allowed:
+            mode: all
+      prefixes:
+      - 100.64.0.1/32
+      evpn:                                    # NEW: EVPN configuration
+        advertiseVNIs: All                     # Advertise all VNIs
+        advertiseSVI: true                     # Advertise SVI IP/MAC
+        l2vnis:                                 # Layer 2 VNI customizations
+        - vni: 1000
+          rd: "64512:1000"
+          importRTs: ["64512:1000"]
+          exportRTs: ["64512:1000"]
+    - asn: 64512
+      vrf: tenant-red
+      prefixes:
+      - 10.0.1.0/24
+      evpn:
+        l3vni:                                  # Layer 3 VNI configuration
+          vni: 2000
+          rd: "64512:2000"
+          importRTs: ["64512:2000", "*:999"]   # Supports wildcard RTs
+          exportRTs: ["64512:2000"]
+          advertisePrefixes: ["unicast"]        # Advertise unicast routes
+```
+
+**New API Types:**
+
+```go
+// Neighbor gets a new field:
+type Neighbor struct {
+    // ... existing fields ...
+
+    // AddressFamilies specifies which address families to activate this neighbor for.
+    // Supported values: "unicast" (IPv4/IPv6 unicast based on neighbor IP), "evpn" (L2VPN EVPN).
+    // Defaults to "unicast" when not specified.
+    // +optional
+    // +kubebuilder:default:=["unicast"]
+    // +kubebuilder:validation:MaxItems=2
+    // +kubebuilder:validation:Enum=unicast;evpn
+    AddressFamilies []string `json:"addressFamilies,omitempty"`
+}
+
+// Router gets a new field:
+type Router struct {
+    // ... existing fields ...
+
+    // EVPN is the EVPN configuration for this router.
+    // +optional
+    EVPN *EVPNConfig `json:"evpn,omitempty"`
+}
+
+// EVPNConfig contains EVPN-specific configuration.
+type EVPNConfig struct {
+    // AdvertiseVNIs controls how VNIs are advertised to EVPN neighbors.
+    // - "Disabled": No VNI advertisements
+    // - "All": Advertise all VNIs (enables advertise-all-vni)
+    // Note: Can only be provided for router instances with EVPN neighbors.
+    // +optional
+    // +kubebuilder:validation:Enum=Disabled;All
+    AdvertiseVNIs *VNIAdvertisement `json:"advertiseVNIs,omitempty"`
+
+    // AdvertiseSVI enables advertising the SVI IP/MAC as a type-2 route.
+    // +optional
+    AdvertiseSVI bool `json:"advertiseSVI,omitempty"`
+
+    // L2VNIs contains configuration for Layer 2 VNIs (MAC-VRF).
+    // Note: Can only be provided for router instances with EVPN neighbors.
+    // +optional
+    L2VNIs []L2VNI `json:"l2vnis,omitempty"`
+
+    // L3VNI contains configuration for the Layer 3 VNI (IP-VRF).
+    // Note: Can only be provided for router instances with no neighbors.
+    // This is a temporary limitation until proper EVPN prefix filtering is implemented.
+    // +optional
+    L3VNI *L3VNI `json:"l3vni,omitempty"`
+}
+
+// VNIAdvertisement defines how VNIs are advertised in EVPN.
+type VNIAdvertisement string
+
+const (
+    VNIAdvertisementDisabled VNIAdvertisement = "Disabled"
+    VNIAdvertisementAll VNIAdvertisement = "All"
+)
+
+// VNI contains common fields for all VNI types.
+type VNI struct {
+    // VNI is the VXLAN Network Identifier (1-16777215).
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:Minimum=1
+    // +kubebuilder:validation:Maximum=16777215
+    VNI uint32 `json:"vni"`
+
+    // RD is the route distinguisher for this VNI.
+    // Format: A.B.C.D:MN|EF:OPQR|GHJK:MN (e.g., "65000:100" or "192.0.2.1:100")
+    // +optional
+    RD RouteDistinguisher `json:"rd,omitempty"`
+
+    // ImportRTs is the list of route targets to import.
+    // Format: A.B.C.D:MN|EF:OPQR|GHJK:MN|*:MN|*:OPQR (e.g., "65000:100", "192.0.2.1:100", "*:100")
+    // +optional
+    // +kubebuilder:validation:MaxItems=100
+    ImportRTs []ImportRouteTarget `json:"importRTs,omitempty"`
+
+    // ExportRTs is the list of route targets to export.
+    // Format: A.B.C.D:MN|EF:OPQR|GHJK:MN (e.g., "65000:100", "192.0.2.1:100")
+    // +optional
+    // +kubebuilder:validation:MaxItems=100
+    ExportRTs []ExportRouteTarget `json:"exportRTs,omitempty"`
+}
+
+// L2VNI represents a Layer 2 VNI configuration (MAC-VRF).
+type L2VNI struct {
+    VNI `json:",inline"`
+}
+
+// L3VNI represents a Layer 3 VNI configuration (IP-VRF).
+type L3VNI struct {
+    VNI `json:",inline"`
+
+    // AdvertisePrefixes controls which prefixes to advertise as EVPN type-5 routes.
+    // - "unicast": advertise the unicast prefixes of the router
+    // +kubebuilder:validation:Required
+    // +kubebuilder:validation:MinItems=1
+    // +kubebuilder:validation:MaxItems=1
+    // +kubebuilder:validation:Enum=unicast
+    AdvertisePrefixes []string `json:"advertisePrefixes"`
+}
+
+// RouteDistinguisher defines an 8-byte BGP identifier.
+// +kubebuilder:validation:XValidation:rule="self.split(':').size() == 2",message="RD must contain exactly one colon"
+// +kubebuilder:validation:XValidation:rule="!self.split(':')[0].contains('.') || (self.split(':')[0].isIP() && uint(self.split(':')[1]) <= 65535u)",message="RD with IPv4 administrator must have format A.B.C.D:MN where MN <= 65535"
+// +kubebuilder:validation:XValidation:rule="self.split(':')[0].contains('.') || uint(self.split(':')[0]) <= 65535u || uint(self.split(':')[1]) <= 65535u",message="RD with 4-byte ASN administrator must have format GHJK:MN where GHJK <= 4294967295 and MN <= 65535"
+// +kubebuilder:validation:XValidation:rule="self.split(':')[0].contains('.') || uint(self.split(':')[0]) > 65535u || uint(self.split(':')[1]) <= 4294967295u",message="RD with 2-byte ASN administrator must have format EF:OPQR where EF <= 65535 and OPQR <= 4294967295"
+type RouteDistinguisher string
+
+// ImportRouteTarget defines a BGP Extended Community for route filtering on import.
+// Supports wildcard matching with "*" as the global administrator (e.g., "*:100").
+// +kubebuilder:validation:XValidation:rule="self.split(':').size() == 2",message="RT must contain exactly one colon"
+// +kubebuilder:validation:XValidation:rule="self.split(':')[0] != '*' || uint(self.split(':')[1]) <= 4294967295u",message="RT with wildcard administrator must have format *:OPQR where OPQR <= 4294967295"
+// +kubebuilder:validation:XValidation:rule="!self.split(':')[0].contains('.') || (self.split(':')[0].isIP() && uint(self.split(':')[1]) <= 65535u)",message="RT with IPv4 administrator must have format A.B.C.D:MN where MN <= 65535"
+// +kubebuilder:validation:XValidation:rule="self.split(':')[0] == '*' || self.split(':')[0].contains('.') || uint(self.split(':')[0]) <= 65535u || uint(self.split(':')[1]) <= 65535u",message="RT with 4-byte ASN administrator must have format GHJK:MN where GHJK <= 4294967295 and MN <= 65535"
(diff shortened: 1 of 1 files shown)


design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+        - vni: 1000
+          rd: "64512:1000"
+          importRTs: ["64512:1000"]
+          exportRTs: ["64512:1000"]
+    - asn: 64512
+      vrf: tenant-red
+      prefixes:
+      - 10.0.1.0/24
+      evpn:
+        l3vni:                                  # Layer 3 VNI configuration
```
fedepaol, line 102: the api allows having a l2vni within the red vrf right? If so, let's add another example

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+metadata:
+  name: evpn-example
+spec:
+  bgp:
+    routers:
+    - asn: 64512
+      neighbors:
+      - address: 192.168.1.1
+        asn: 64512
+        addressFamilies: ["unicast", "evpn"]  # NEW: activate EVPN address family
```
fedepaol: this allows to configure unicast only neighbors. Let's add another example so it's clear(er)

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+- Support both Layer 2 VNI (MAC-VRF, bridging) and Layer 3 VNI (IP-VRF, routing) configuration
+- Enable BGP neighbor activation for the L2VPN EVPN address family
+- Support route distinguisher (RD) and route target (RT) configuration for VNIs
+
+### Non-Goals
+
+The following features are not required by OVN-Kubernetes and are therefore out of scope for this proposal:
+
+- **Route-map based filtering and customization**: Advanced route filtering and customization (prefix filtering, VNI filtering, route modification) via route-maps
+- **Advanced EVPN features**: Multi-homing, BUM handling customization, advertise-PIP, advertise-svi-ip, and other advanced EVPN capabilities beyond basic L2/L3 VPN functionality
```
fedepaol, line 37: I'd add advertise-svi-ip even if it's not required by ovnk

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+
+## Proposal
+
+### User Stories
+
+As a **cluster administrator**, I want to:
+
+1. **Configure EVPN neighbors**: Enable BGP sessions with external EVPN routers (route reflectors or VTEPs) by activating the L2VPN EVPN address family on existing BGP neighbors.
+
+2. **Advertise L2 VNIs**: Advertise L2 VNIs to EVPN neighbors, enabling Layer 2 connectivity to external networks.
```
fedepaol: just terminology, but this is more "advertise type 5 routes / type 3 routes" associated to l2 / l3 vnis

fedepaol: also, I guess we also want to receive them

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+type Neighbor struct {
+    // ... existing fields ...
+
+    // AddressFamilies specifies which address families to activate this
+    // neighbor for. Supported values: "unicast", "evpn".
+    // +optional
+    // +kubebuilder:default:={"unicast"}
+    // +kubebuilder:validation:MaxItems=2
+    // +kubebuilder:validation:Enum=unicast;evpn
+    AddressFamilies []string `json:"addressFamilies,omitempty"`
```
fedepaol, line 124: will the toAdvertise / toReceive filter impact the evpn address family too? (I think we should have a way to select what prefixes we want)

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+}
+
+// EVPNConfig contains EVPN-specific configuration.
+type EVPNConfig struct {
+    // AdvertiseVNIs controls how VNIs are advertised to EVPN neighbors.
+    // - "Disabled": No VNI advertisements
+    // - "All": Advertise all VNIs (enables advertise-all-vni)
+    // Note: Can only be provided for router instances with EVPN neighbors.
+    // +optional
+    // +kubebuilder:validation:Enum=Disabled;All
```
fedepaol, line 143: how about having this "advertiseAllVNIs" as a bool? I don't think there's room for new values as it matches what frr offers

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+// VNIAdvertisement defines how VNIs are advertised in EVPN.
+type VNIAdvertisement string
+
+const (
+    VNIAdvertisementDisabled VNIAdvertisement = "Disabled"
+    VNIAdvertisementAll VNIAdvertisement = "All"
+)
+
+// VNIBase contains common fields for all VNI types.
+type VNIBase struct {
```
fedepaol: I'd just call this "VNI"

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+6. **Configuration updates**:
+   - Add/remove VNIs dynamically
+   - Change RD/RT values
+   - Verify FRR configuration is updated correctly
+
+7. **Negative tests**:
+   - Duplicate VNI numbers (should be rejected)
+   - Invalid RD/RT formats (should be rejected)
+   - Invalid configuration in non underlay EVPN routers (should be rejected)
+
```
fedepaol, line 438: given the host will be configured properly, I think having some basic traffic test would be useful

fedepaol, line 438: we should test a mixed scenario where some neighbors are configured for bgp and some others for evpn

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+### Phase 5: Integration Testing
+- Set up EVPN-capable external FRR containers
+- Implement integration tests covering all use cases
+- Verify EVPN session establishment
+- Verify route advertisement and filtering
+
+### Phase 6: Documentation
+- Update API documentation
+- Add EVPN configuration examples
+- Document OVN-Kubernetes integration patterns
```
fedepaol: I'd just add a reference to ovnk

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+### Phase 4: Template Rendering
+- Create `internal/frr/templates/evpn.tmpl`
+- Update `internal/frr/templates/frr.tmpl` to include EVPN template
+- Handle address family activation for neighbors
+- Handle VRF declaration with VNI binding
+- Handle L2 VNI blocks in EVPN address family
+- Handle L3 VNI RD/RT in EVPN address family (no vni block)
+- Add unit tests with golden file comparison
+
+### Phase 5: Integration Testing
```
fedepaol: somewhere between here and `#6` we should have a raw implementation of the host configuration that can be consumed by any project which wants to leverage this new feature (as I suggested in https://github.com/metallb/frr-k8s/pull/372#issuecomment-3681070091 ). Let's try to shape it up too before the implementation

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+- Enable BGP neighbor activation for the L2VPN EVPN address family
+- Support route distinguisher (RD) and route target (RT) configuration for VNIs
+
+### Non-Goals
+
+The following features are not required by OVN-Kubernetes and are therefore out of scope for this proposal:
+
+- **Route-map based filtering and customization**: Advanced route filtering and customization (prefix filtering, VNI filtering, route modification) via route-maps
+- **Advanced EVPN features**: Multi-homing, BUM handling customization, advertise-PIP, advertise-svi-ip, and other advanced EVPN capabilities beyond basic L2/L3 VPN functionality
+- **Alternative data planes**: This proposal only supports EVPN over VXLAN. Other data planes (MPLS, SR-MPLS, SRv6) are out of scope
```
fedepaol, line 38: let's state explicitly that configuring the host is a non goal, and this will require host configuration either manually or via some other controller

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+   - L3VNI prefix advertisement rendering
+   - VRF declaration with VNI binding
+
+#### Integration Tests
+
+Using the existing external FRR containers infrastructure:
+
+1. **Basic EVPN session establishment**:
+   - Configure FRRConfiguration with EVPN neighbor
+   - Verify EVPN session comes up
```
fedepaol: BGP session?

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+- The `addressFamilies` field defaults to `["unicast"]`, maintaining current behavior
+
+#### Validation
+
+Admission webhooks will enforce:
+
+1. **Unique VNI numbers**: VNI numbers must be unique across all routers in a FRRConfiguration
+2. **Scope validation**: `advertiseVNIs` and `l2vnis` can only be configured on routers with EVPN neighbors
+
+#### Metrics
```
fedepaol, line 353: can we add a section about the merging logic and what's going to be considered "mergeable" and how?

design/basic-evpn-support.md
```diff
@@ -0,0 +1,395 @@
...
+
+**New API Types:**
+
+```go
+// Neighbor gets a new field:
+type Neighbor struct {
+    // ... existing fields ...
+
+    // AddressFamilies specifies which address families to activate this
+    // neighbor for. Supported values: "unicast", "evpn".
```
fedepaol: I am a bit torn about this and the interaction with dualStackAddressFamily, as we have two places to change to configure address family. I'd be tempted to call this v1beta2 and have a conversion webhook between the old flag and this. Wdyt?

jcaamano: My idea is that `dualStackAddressFamily` only applies to unicast but other than documenting it I see how that is confusing.
First, is that your idea as well?
Second, I guess you see potential values here `ipv4`, `ipv6` and `evpn` or something of the sort?
In regards to `v1beta2`, that's one option. But given that the API is still already beta anyways, why not deprecate `dualStackAddressFamily` directly?

design/basic-evpn-support.md
jcaamano: Is there a reference to other design proposal already covering this? Just want to have some guidance on how this was approached before if available.

design/basic-evpn-support.md
jcaamano: What do you mean with "before the implementation"?

design/basic-evpn-support.md
jcaamano: I envision here a future potential value where only the specified VNIs are advertised: so for example instead of having FRR picking up and advertising all L2 VNIS configured in the host, just the L2 VNIs that are explicitly include in this configuration are advertised.

design/basic-evpn-support.md
jcaamano: No, it wouldn't but I agree it's reasonable to think we should and this is a discussion I wanted to have.

Let's focus first on how we would accommodate it in the API. I have two ideas:
1. Additional fields in/for toAdvertise/toRecieve that clarify what address family they apply to
2. Instead of specifying a neighbor in multiple address families, specify it for each address family (potentially specifying the same neighbor twice if you need to configure it for both; not great, but not terrible given that frr-k8s is already in the business of merging).

Would you have other ideas?

Additionally, for EVPN we can have filtering on the IPVRF level so we can expect something like a toAdvertise in L3VNI as well.

Then we would need to discuss if we want to implement this now. OVNK does not have a use for it and it wouldn't be trivial.

design/basic-evpn-support.md
jcaamano:
```
    // L2VNIs contains configuration for Layer 2 VNIs (MAC-VRF).
    // Note: Can only be provided for router instances with EVPN neighbors.
```
So in this example we wouldn't be able to have a l2vni in tenant-red. 

But I guess we could define a router in a VRF with EVPN neighbors acting as the underlay. Is that the example  you are after?

fedepaol: what I mean is having a l2 vni having a l3 vni as routing domain (basically, what's in the example here https://docs.frrouting.org/en/latest/evpn.html#traditional-bridges-and-traditional-vxlan-devices for vni110, where vrf1 - vni 100 is the routing domain for vni110).

The underlay still belongs to the default router / vrf

design/basic-evpn-support.md
fedepaol: > My idea is that `dualStackAddressFamily` only applies to unicast but other than documenting it I see how that is confusing. First, is that your idea as well? Second, I guess you see potential values here `ipv4`, `ipv6` and `evpn` or something of the sort? In regards to `v1beta2`, that's one option. 

`ipv4`, `ipv6`, `evpn` or even `dual`, to be discussed

> But given that the API is still already beta anyways, why not deprecate `dualStackAddressFamily` directly?

wearing my red - hat here :sweat_smile:  it's beta but supported in openshift, so I'd try to be careful in deprecating

design/basic-evpn-support.md
fedepaol: > No, it wouldn't but I agree it's reasonable to think we should and this is a discussion I wanted to have.
> 
> Let's focus first on how we would accommodate it in the API. I have two ideas:
> 
>     1. Additional fields in/for toAdvertise/toRecieve that clarify what address family they apply to
> 
>     2. Instead of specifying a neighbor in multiple address families, specify it for each address family (potentially specifying the same neighbor twice if you need to configure it for both; not great, but not terrible given that frr-k8s is already in the business of merging).
> 
> 
> Would you have other ideas?
>

We need to clarify what we want to do first, and I see a couple of options:

1- we can choose to filter at ipvrf level, regardless of the neighbor. This might make sense as I don't expect a need to differentiate what prefixes are being sent depending on the other end of the underlay. 
In this scenario, what is being filtered (as you write below) belongs more to L3VNI rather than the neighbor, as we might want to say "I want to advertise / receive this prefix to vni 300".

2- we can choose to filter at l3 vni level PLUS neighbor, meaning we want a way to express something like "I want to advertise / receive this prefix to vni 300 for neighbor 1.2.3.4".

I think the first is more than enough. After that, we need to document the fact that toAdvertise / toReceive have effect only on BGP routes.

> Additionally, for EVPN we can have filtering on the IPVRF level so we can expect something like a toAdvertise in L3VNI as well.
> 
> Then we would need to discuss if we want to implement this now. OVNK does not have a use for it and it wouldn't be trivial.

I won't block on this assuming we have a path forward which will be implemented later, but we must care to ensure plain BGP and EVPN won't interfere.

When you have a neighbor in a VRF, there's a risk that the prefixes being advertised / received to BGP will have an effect to EVPN too (because those prefixes end up in the network statement). One possible option is (as a stop gap solution until we have proper filtering) to block defining a neighbor in a VRF if that VRF is used for EVPN. Alternatively, we can check if moving the network xxxx statement under the l2vpn evpn address family will make the trick, so you could say "unicast" but also a very specific prefix.

design/basic-evpn-support.md
fedepaol: nope, mostly because merging was there from the beginning. Let's try to put together a list of fields that will be considered mergeable and what not.

design/basic-evpn-support.md
fedepaol: sorry, I meant before starting to implement the e2e tests

design/basic-evpn-support.md
jcaamano: I think I understand what you mean, but I don't see how that can is reflected in an FRR configuration (like the one I am sampling here). If you actually look at the FRR Configuration from the example you are bringing up (https://docs.frrouting.org/en/latest/evpn.html#sample-configuration) that's not reflected there in any way; or is there something I am missing?

fedepaol: ah, you are right. This is because the discriminator is the VRF the bridge belongs to :sweat_smile: 
Ignore my comment then

design/basic-evpn-support.md
jcaamano: > wearing my red - hat here 😅 it's beta but supported in openshift, so I'd try to be careful in deprecating

My 2 cents here: deprecating is not removing support and I think that moving to a new API version is best to happen at logical stages of the project when it is actually a good time to remove deprecated fields. If you don't do deprecate-then-remove, and you just remove directly then you might end up with a lot of API versions, don't you? Like if you had v1beta2 and then wanted to deprecate another field and you immediately did v1beta3 and so on...

Anyway, I am ok with doing v1beta2 if you think this is a good time!

design/basic-evpn-support.md
jcaamano: > We need to clarify what we want to do first, and I see a couple of options:
> 
> 1- we can choose to filter at ipvrf level, regardless of the neighbor. This might make sense as I don't expect a need to differentiate what prefixes are being sent depending on the other end of the underlay. In this scenario, what is being filtered (as you write below) belongs more to L3VNI rather than the neighbor, as we might want to say "I want to advertise / receive this prefix to vni 300".
> 
> 2- we can choose to filter at l3 vni level PLUS neighbor, meaning we want a way to express something like "I want to advertise / receive this prefix to vni 300 for neighbor 1.2.3.4".
> 
> I think the first is more than enough. After that, we need to document the fact that toAdvertise / toReceive have effect only on BGP routes.

I think we need to expect option 2 happening eventually. And if I were to put an order to it, I think neighbor level filtering provides the most capability: IPVRF level filtering just works in one direction: unicast -> evpn, doesn't it? so it can't filter on receive? Not sure if I am missing something.

> When you have a neighbor in a VRF, there's a risk that the prefixes being advertised / received to BGP will have an effect to EVPN too (because those prefixes end up in the network statement). One possible option is (as a stop gap solution until we have proper filtering) to block defining a neighbor in a VRF if that VRF is used for EVPN. Alternatively, we can check if moving the network xxxx statement under the l2vpn evpn address family will make the trick, so you could say "unicast" but also a very specific prefix.

But that would be a result of an explicit configuration, wouldn't it? A user would have the option to not configure evpn and unicast neighbors in the same router it if it does not want that to happen, versus us blocking a potential desired configuration while not really providing anything of value other than an interpretation of what might or might not be a mis-configuration?

design/basic-evpn-support.md
fedepaol: iiuc, your suggestion is to deprecate the old field, implement the new logic in the same api version and then create a new api version when things are stable enough and we want to remove the field (but still support the old api version for a while, I guess?)

In any case, yes just deprecating (but keeping it working) and implementing the new behavior seems more sane to me, thanks!

design/basic-evpn-support.md
fedepaol: > > We need to clarify what we want to do first, and I see a couple of options:
> > 1- we can choose to filter at ipvrf level, regardless of the neighbor. This might make sense as I don't expect a need to differentiate what prefixes are being sent depending on the other end of the underlay. In this scenario, what is being filtered (as you write below) belongs more to L3VNI rather than the neighbor, as we might want to say "I want to advertise / receive this prefix to vni 300".
> > 2- we can choose to filter at l3 vni level PLUS neighbor, meaning we want a way to express something like "I want to advertise / receive this prefix to vni 300 for neighbor 1.2.3.4".
> > I think the first is more than enough. After that, we need to document the fact that toAdvertise / toReceive have effect only on BGP routes.
> 
> I think we need to expect option 2 happening eventually. And if I were to put an order to it, I think neighbor level filtering provides the most capability: IPVRF level filtering just works in one direction: unicast -> evpn, doesn't it? so it can't filter on receive? Not sure if I am missing something.
> 

I thought it'd made more sense to filter per VNI rather than per neighbor + vni, as one might be interested in receiving / propagating prefixes to the virtual network. In any case, I think the direction the api is taking doesn't prevent it, so we can postpone the discussion when we get to it.


> > When you have a neighbor in a VRF, there's a risk that the prefixes being advertised / received to BGP will have an effect to EVPN too (because those prefixes end up in the network statement). One possible option is (as a stop gap solution until we have proper filtering) to block defining a neighbor in a VRF if that VRF is used for EVPN. Alternatively, we can check if moving the network xxxx statement under the l2vpn evpn address family will make the trick, so you could say "unicast" but also a very specific prefix.
> 
> But that would be a result of an explicit configuration, wouldn't it? A user would have the option to not configure evpn and unicast neighbors in the same router it if it does not want that to happen, versus us blocking a potential desired configuration while not really providing anything of value other than an interpretation of what might or might not be a mis-configuration?

The configuration would be explicit, but given the way merge happens, nothing prevents having a scenario where for the same vrf:

- one side configures a bgp session with a neighbor, with the intent to advertise (and maybe receive) prefixes
- another side configures EVPN with the same vrf and some prefixes to advertise

What would happen is that the prefixes get merged all together (because they belong to the router) and advertised both via BGP and via EVPN. 

So, the configuration is legit but the end result is probably not what the user meant.

jcaamano: > I thought it'd made more sense to filter per VNI rather than per neighbor + vni
> In any case, I think the direction the api is taking doesn't prevent it, so we can postpone the discussion when we get to it.

I probably was not understanding. Do you mean defining the filtering by expanding, in the future, the definition of the L2VNI and L3VNI structs I am proposing here?

I had thought of that too, not sure why I forgot to mention it before. That's ok too, with the risk of being to repetitive by means of having similar toAdvertise/toReceieve fields in different parts of the API.

design/basic-evpn-support.md
jcaamano: For the sake of completeness, there's the option of not deprecating/removing. So we could just use `unicast` value meaning "enable the neighbor not only for the ipfamily related to its session" and still use `dualStackAddressFamily` to also enable the other unicast ipfamily.

I see how if we were starting from scratch, your suggestion makes more sense, but this other consideration might be valuable to avoid throwing off users with a deprecation.

fedepaol: ah that's correct, and that was the original proposal I think. Let's keep things simple for now and stick to that (meaning, `unicast` looks at `dualstackaddressfamily`). We can optionally add a `unicast - v4` and `unicast - v6` which prepares the api to deprecate `dualstackaddressfamily` and the bare `unicast` value.

design/basic-evpn-support.md
fedepaol: > > I thought it'd made more sense to filter per VNI rather than per neighbor + vni
> > In any case, I think the direction the api is taking doesn't prevent it, so we can postpone the discussion when we get to it.
> 
> I probably was not understanding. Do you mean defining the filtering by expanding, in the future, the definition of the L2VNI and L3VNI structs I am proposing here?

Yep

> 
> I had thought of that too, not sure why I forgot to mention it before. That's ok too, with the risk of being to repetitive by means of having similar toAdvertise/toReceieve fields in different parts of the API.

Right, but those have different targets so even if the fields are similar, the filters are probably going to be different

design/basic-evpn-support.md
```diff
@@ -0,0 +1,494 @@
...
+    // +kubebuilder:validation:Maximum=16777215
+    VNI uint32 `json:"vni"`
+
+    // RD is the route distinguisher for this VNI.
+    // Format: A.B.C.D:MN|EF:OPQR|GHJK:MN (e.g., "65000:100" or "192.0.2.1:100")
+    // +optional
+    RD RouteDistinguisher `json:"rd,omitempty"`
+
+    // ImportRTs is the list of route targets to import.
+    // Format: A.B.C.D:MN|EF:OPQR|GHJK:MN|*:OPQR|*:MN (e.g., "65000:100" or "192.0.2.1:100" or "*:42000000")
```
fedepaol: you probably checked already, but are you sure frr supports wildcarding in route targets (same for export)?

design/basic-evpn-support.md
```diff
@@ -0,0 +1,494 @@
...
+    // +optional
+    // +kubebuilder:validation:MaxItems=100
+    ImportRTs []RouteTarget `json:"importRTs,omitempty"`
+
+    // ExportRTs is the list of route targets to export.
+    // Format: A.B.C.D:MN|EF:OPQR|GHJK:MN|*:OPQR|*:MN (e.g., "65000:100" or "192.0.2.1:100" or "*:42000000")
```
fedepaol: what's the point of wildcarding an exported route target?

jcaamano: None! tweaked accordingly.

design/basic-evpn-support.md
jcaamano: [It supposedly does](https://docs.frrouting.org/en/latest/bgp.html#evpn-l3-route-targets):
```
EVPN L3 Route-Targets

route-target <import|export|both> <RTLIST|auto>

Modify the route-target set for EVPN advertised type-2/type-5 routes. RTLIST is a list of any of matching 
(A.B.C.D:MN|EF:OPQR|GHJK:MN|*:OPQR|*:MN) where * indicates wildcard matching for the AS number. It will be 
set to match any AS number. This is useful in datacenter deployments with Downstream VNI. auto is used to 
retain the autoconfigure that is default behavior for L3 RTs.
```

design/basic-evpn-support.md
```diff
@@ -0,0 +1,504 @@
...
+**Mergeable fields** (values are aggregated):
+
+1. **ImportRTs**: Import route targets for the same VNI are aggregated from all FRRConfigurations and duplicates are removed
+2. **ExportRTs**: Export route targets for the same VNI are aggregated from all FRRConfigurations and duplicates are removed
+3. **AdvertisePrefixes** (L3VNI only): Values are aggregated from all FRRConfigurations and duplicates are removed
+4. **L2VNIs/L3VNI**: Different VNI numbers can be contributed by different FRRConfigurations, but each unique VNI number must follow the rules above for its per-VNI fields
+
+This merging behavior allows users to split EVPN configuration across multiple FRRConfiguration resources while maintaining consistency. For example:
+
+- One FRRConfiguration can set `advertiseVNIs: All` for a router
```
fedepaol: isn't this in contrast with "advertiseVNIs must be teh same" above (I think this makes more sense)?

design/basic-evpn-support.md
```diff
@@ -0,0 +1,504 @@
...
+
+**Non-mergeable fields** (conflicting values cause validation failure):
+
+1. **AdvertiseVNIs**: If multiple FRRConfigurations specify different `advertiseVNIs` values for the same router, the configuration is invalid
+2. **AdvertiseSVI**: If multiple FRRConfigurations specify different `advertiseSVI` values for the same router, the configuration is invalid
+3. **RD (Route Distinguisher)**: If the same VNI number appears in multiple FRRConfigurations with different RD values, the configuration is invalid
+
+**Mergeable fields** (values are aggregated):
+
+1. **ImportRTs**: Import route targets for the same VNI are aggregated from all FRRConfigurations and duplicates are removed
```
fedepaol: How do we want to treat two frrconfigurations, one with RT and the other without it? I think they are conflicting unless there's a way to say "I want also the default RT", because with the specific RT we are effectively overriding the default one (and maybe breaking a working frrconfiguration).

It's a bit different from merging two configs with different explicit route targets imo, as those will be added together.

design/basic-evpn-support.md
jcaamano: It is in contrast. Yes, `advertiseVNIs must be the same"` because if not they are contradicting and we shouldn't give more weight to one value vs another.

jcaamano: Will clarify

design/basic-evpn-support.md
jcaamano: Yeah, this is something worth clarifying explicitly.

We can either:
- The resulting RT is the union of all RTs. if that union ends up being empty, then default RT applies.
- We consider merging empty with non empty RT a conflict.

I was implying the former option. We can go for the latter if you think it fits best. Will clarify either way.

fedepaol: Yep, what I find disturbing here is that empty means "use the default", so we could potentially have one configuration breaking the other because it overrides the default (unless there's a way to express "default RT plus this one", but I think it's still better to reject merging empty vs non empty)

PR #389, opened 2026-01-09
Fix build make target

andreaskaris:
This commit fixes the `build` Make target. The target isn't really used
anywhere, but can be used for sanity checks for quick builds on the
local system without Docker / podman: Instead of only building frr-k8s,
also build the other 3 binaries (frr-metrics, frr-status,
statuscleaner).

/kind cleanup

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,10 @@ test: manifests generate fmt vet envtest ## Run tests.
 
 .PHONY: build
 build: manifests generate fmt vet ## Build k8s-frr binary.
-	go build -o bin/frr-k8s cmd/main.go
+	go build -v -o bin/frr-k8s ./cmd/frr-k8s-controller
+	go build -v -o bin/frr-metrics ./cmd/metrics
+	go build -v -o bin/frr-status ./cmd/status
+	go build -v -o bin/statuscleaner ./cmd/statuscleaner
 
 .PHONY: run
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -75,7 +75,10 @@ test: manifests generate fmt vet envtest ## Run tests.
 
 .PHONY: build
 build: manifests generate fmt vet ## Build k8s-frr binary.
-	go build -o bin/frr-k8s cmd/main.go
+	go build -o bin/frr-k8s cmd/frr-k8s-controller/main.go
```
fedepaol: since we are here, let's transition to using the folder as a target:

```go
go build -o bin/frr-k8s cmd/frr-k8s-controller
```

This will allow the build to succeed in case we'll add files under those dirs (otherwise it'd compile only main.go, complaining about unresolved deps, been there :sweat_smile: )

andreaskaris: done, needed a leading ./ though

andreaskaris:
> what exactly is the failure and where when that commit is not merged?

I updated the commit message. It's just that the 3 other binaries are missing from the build, so technically it's just a minor improvement to build all of them manually, outside of the container

PR #419, opened 2026-03-18
Implementation of the EVPN design proposal 

jcaamano:
Implements the EVPN design proposal from #372 (design/basic-evpn-support.md).                                            

- [x] API types and CEL validation
- [x] Internal types and translation
- [x] Configuration merging
- [x] FRR template rendering
- [x] Unit and controller tests
- [x] E2e tests (L2 VNI, L3 VNI, composition)
- [x] README and host setup script

Deviations from the design proposal                                                                                      
  - MaxItems constraints for CEL cost budget: Several fields required MaxItems limits to stay within the Kubernetes   CEL validation cost budget:
    - bgp.routers: 50 (pre-existing field — action required for users with more than 50 routers)                           
    - evpn.l2vnis: 100                                                                                                                                                                                  
    - importRTs, exportRTs: 10 each                                                                                        


/kind feature

```release-note
  Add basic EVPN support to FRRConfiguration. Users can now configure BGP neighbors for the L2VPN EVPN address family and  
  advertise Layer 2 VNIs (type-2/type-3 routes) and Layer 3 VNIs (type-5 prefix routes) with configurable route            
  distinguishers and route targets.                                                                                        
                                                                                                                           
  Action required: A MaxItems=50 validation was added to the `bgp.routers` field to comply with the CEL cost budget.       
  Configurations with more than 50 routers must be split across multiple FRRConfiguration resources.
```

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>


jcaamano:
tested with OVNK here
https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6127

which moves OVNK to the API and its evpn lane passes succesfully

fedepaol:
Couple of initial comments: please add more color to the commit messages, and fix the conflicts

api/v1beta1/frrconfiguration_types.go
```diff
@@ -413,3 +426,135 @@ const (
...
+type RouteDistinguisher string
+
+// ImportRouteTarget defines a BGP Extended Community for route filtering on import.
+// Supports wildcard matching with "*" as the global administrator (e.g., "*:100").
+// +kubebuilder:validation:MaxLength=21
+// +kubebuilder:validation:XValidation:rule="self.split(':').size() == 2",message="RT must contain exactly one colon"
+// +kubebuilder:validation:XValidation:rule="self.split(':').size() != 2 || (self.startsWith('*:') || isIP(self.split(':')[0]) || self.split(':')[0].matches('[0-9]+'))",message="RT global administrator must be either '*', an IPv4 address, or a number"
+// +kubebuilder:validation:XValidation:rule="self.split(':').size() != 2 || self.split(':')[1].matches('[0-9]+')",message="RT local administrator must be a number"
+// +kubebuilder:validation:XValidation:rule="self.split(':').size() != 2 || !self.startsWith('*:') || (self.split(':')[1].matches('[0-9]+') && uint(self.split(':')[1]) <= 4294967295u)",message="RT with wildcard global administrator must have format *:OPQR where OPQR <= 4294967295"
```
fedepaol, line 545: general comment about these: do we have a way to test them? Might be me, but it feels hard to validate them otherwise

jcaamano: tested in `internal/controller/frrconfiguration_crd_validation_test.go`

fedepaol: ah awesome!

jcaamano:
Had to pin the envtest version used from `latest` to `release-20`, otherwise:
```
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20260405152528-6210f847b2c1
go: sigs.k8s.io/controller-runtime/tools/setup-envtest@latest: sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20260405152528-6210f847b2c1 requires go >= 1.26.0 (running go 1.25.8; GOTOOLCHAIN=local)
```

Maybe its preferable to bump to go 1.26?

jcaamano:
I don't think the `[CI / metallb_e2e (pull_request)](https://github.com/metallb/frr-k8s/actions/runs/24185452294/job/70589144963?pr=419)` issues are related to this PR as it is failing in the same way for other PRs

internal/controller/api_to_config.go
```diff
@@ -599,26 +610,18 @@ func bfdProfileToFRR(bfdProfile v1beta1.BFDProfile) *frr.BFDProfile {
...
 
-func sortMapPtr[T any](toSort map[string]*T) []T {
-	keys := make([]string, 0)
-	for k := range toSort {
-		keys = append(keys, k)
-	}
-	sort.Strings(keys)
-	res := make([]T, 0)
+func sortMapPtr[K cmp.Ordered, T any](toSort map[K]*T) []T {
+	keys := slices.Sorted(maps.Keys(toSort))
```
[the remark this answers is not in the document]
jcaamano: The project go.mod is targeting 1.25: https://github.com/metallb/frr-k8s/blob/main/go.mod#L3-L5

internal/controller/api_to_config.go
```diff
@@ -599,26 +610,18 @@ func bfdProfileToFRR(bfdProfile v1beta1.BFDProfile) *frr.BFDProfile {
...
 
-func sortMap[T any](toSort map[string]T) []T {
-	keys := make([]string, 0)
-	for k := range toSort {
-		keys = append(keys, k)
-	}
-	sort.Strings(keys)
-	res := make([]T, 0)
+func sortMap[K cmp.Ordered, T any](toSort map[K]T) []T {
+	keys := slices.Sorted(maps.Keys(toSort))
```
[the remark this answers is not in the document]
jcaamano: The project go.mod is targeting 1.25: https://github.com/metallb/frr-k8s/blob/main/go.mod#L3-L5

api/v1beta1/frrconfiguration_types.go
```diff
@@ -59,6 +59,7 @@ type RawConfig struct {
 type BGPConfig struct {
 	// Routers is the list of routers we want FRR to configure (one per VRF).
 	// +optional
+	// +kubebuilder:validation:MaxItems=50
```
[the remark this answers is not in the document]
jcaamano: As noted in the release notes.

internal/controller/api_to_config.go
```diff
@@ -730,3 +733,113 @@ func validateRouterConfig(r *frr.RouterConfig) error {
...
+
+	hasEVPNNeighbor := false
+	for _, n := range r.Neighbors {
+		if slices.Contains(n.AddressFamilies, string(v1beta1.AddressFamilyEVPN)) {
+			hasEVPNNeighbor = true
+			break
+		}
+	}
+
+	needsEVPNNeighbor := r.EVPN.AdvertiseVNIs != nil || r.EVPN.AdvertiseSVI || len(r.EVPN.L2VNIs) > 0
```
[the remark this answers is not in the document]
jcaamano: @fedepaol I guess this makes sense, WDYT?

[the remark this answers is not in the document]
jcaamano: Or the other option is remove the `Disabled` enum value. I guess I just added it because it made sense to have an explicit value corresponding to the implicit default.

jcaamano:
Also we need to think about metrics. My first thought is equivalent metrics to unicast where applicable.

api/v1beta1/frrconfiguration_types.go
fedepaol, line 62: what's the rationale behind this limit?

api/v1beta1/frrconfiguration_types.go
```diff
@@ -413,3 +426,135 @@ const (
...
+	VNIAdvertisementAll VNIAdvertisement = "All"
+)
+
+// VNI contains common fields for all VNI types.
+type VNI struct {
+	// VNI is the VXLAN Network Identifier (1-16777215).
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=16777215
+	VNI uint32 `json:"vni"`
```
fedepaol: just realized we'll find ourselves using VNI.VNI which doesn't read beautifully.

One possible option is to move VNI one level up (duplicating in l3vni and l2vni) and call this Attributes / Properties

internal/controller/api_to_config.go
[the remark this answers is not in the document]
fedepaol: I think the suggestion is fine. One thing I'd do though is to move the gigantic if in a function as it's becoming too crowded

internal/frr/templates/neighboripfamily.tmpl
```diff
@@ -5,14 +5,14 @@
   {{- $peer = .Iface }}
 {{- end }}
 {{/* no bgp default ipv4-unicast prevents peering if no address families are defined. We declare an ipv4 one for the peer to make the pairing happen */}}
-{{- if activateNeighborFor "ipv4" .IPFamily }}
+{{- if and (hasAddressFamilyUnicast .AddressFamilies) (activateNeighborFor "ipv4" .IPFamily) }}
```
fedepaol: let's embed this logic in "activateNeighbor" by providing addressfamilies to the function

fedepaol (commented):
Reviewed up to e2e tests excluded, few minor comments

api/v1beta1/frrconfiguration_types.go
jcaamano: Budget allows ~6X routers, rounded to 50

internal/controller/api_to_config.go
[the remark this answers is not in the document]
jcaamano: AI's suggestion I guess you mean?

api/v1beta1/frrconfiguration_types.go
jcaamano: Done. Please note the weirdness in the API-DOCs file (embed fields are docummented as appearing in both in embedded struct and the struct you are embedding in).

internal/frr/templates/neighboripfamily.tmpl
jcaamano: I though it was cleaner my way of keeping them separate but I took your queue. If you reconsider, `activateNeighborFor` could perhaps be renamed to something more clear as `isNeighborIPFamily`

jcaamano:
For now, rework on its own commits for ease of review. Will squash before merge or upon request.

jcaamano:
The failure `[FAIL] Unnumbered BGP session is established and routes are verified [It] eBGP` in the dual stack lane looks like a flake

internal/controller/api_to_config.go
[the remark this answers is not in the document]
fedepaol: yep, AI's suggestoin

fedepaol:
> Also we need to think about metrics. My first thought is equivalent metrics to unicast where applicable.

I think we can handle it in a separate PR to avoid having this too crowded

api/v1beta1/frrconfiguration_types.go
fedepaol: not sure I understood what budget is, mind expanding?

internal/controller/api_to_config.go
```diff
@@ -730,3 +733,115 @@ func validateRouterConfig(r *frr.RouterConfig) error {
...
+
+	hasEVPNNeighbor := false
+	for _, n := range r.Neighbors {
+		if slices.Contains(n.AddressFamilies, string(v1beta1.AddressFamilyEVPN)) {
+			hasEVPNNeighbor = true
+			break
+		}
+	}
+
+	hasAdvertiseAll := r.EVPN.AdvertiseVNIs != nil && *r.EVPN.AdvertiseVNIs == string(v1beta1.VNIAdvertisementAll)
```
fedepaol: I'd move this lenghty ifs to functions for better readability

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,755 @@
...
+	evpnVxlan      = "evpnvx"
+	evpnL3VRFTable = 10
+	// evpnAccessIface is the name of the veth interface created by hack/evpn-setup.sh
+	// on each container for L2 VNI local connectivity. Its MAC is advertised as a
+	// type-2 route via EVPN.
+	evpnAccessIface = "evpnacc"
+)
+
+// runEVPNSetupScript runs hack/evpn-setup.sh with the given parameters.
+func runEVPNSetupScript(nodes []corev1.Node, externalFRR *frrcontainer.FRR, l2 bool, l3 bool, cleanup bool) error {
```
fedepaol: here we are taking a mixed approach where we run the script on the external frr as part of the e2e tests, but we run it on the kind nodes outside of the test run.
I'd keep it uniform.
Also, if we want to be able to recycle these tests in Openshift CI, imo it makes sense to use an approach where we copy the scripts to the frrk8s pods and we run them from there (I think it should be doable as the pods are hostnetworked and super powered)

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,755 @@
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("evpn-setup.sh failed: %w\noutput: %s", err, string(out))
+	}
+	return nil
+}
+
+// evpnNeighborHasAddressFamily checks that a neighbor on the given executor
+// has the l2vpn evpn address family active.
+func evpnNeighborHasAddressFamily(exec executor.Executor, neighborIP string) error {
```
fedepaol: I think it makes sense to move all these evpn* functions to an evpn package

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,755 @@
...
+		if !found {
+			return fmt.Errorf("type-5 route %s expected next-hop %s, got %v", routeKey, expectedNH, nhs)
+		}
+	}
+	return nil
+}
+
+// evpnPing pings the target IP from the given container and returns an error
+// if the ping fails. Used to validate EVPN data path connectivity.
+func evpnPing(container, targetIP string) error {
```
fedepaol: this is not related to evpn, might belong to its own package (or otherwise I'd rename it just to "ping" and keep it local)

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,755 @@
...
+	out, err := executor.Host.Exec(executor.ContainerRuntime, "exec", container,
+		"ping", "-c", "1", "-W", "2", targetIP)
+	if err != nil {
+		return fmt.Errorf("ping %s from %s failed: %w\noutput: %s", targetIP, container, err, out)
+	}
+	return nil
+}
+
+// evpnPingVRF pings the target IP from within a VRF on the given container.
+func evpnPingVRF(container, vrf, targetIP string) error {
```
fedepaol: in general, if we have util functions left after moving to a package, I'd move them to the bottom of the file

e2etests/tests/testdata/evpn-setup.sh
```diff
@@ -0,0 +1,530 @@
...
+# Creates a bridge with VLAN filtering and a VXLAN device in external/vnifilter
+# mode. The VXLAN device supports multiple VNIs on a single device; FRR manages
+# VNI-to-VTEP mappings via BGP EVPN signaling.
+#
+# Args: $1=container, $2=local VTEP IP
+setup_bridge_vxlan() {
+    local container=$1 local_ip=$2
+
+    log "[$container] Creating bridge and VXLAN (local VTEP: $local_ip)..."
+    exec_in "$container" "
```
fedepaol: see my comment on test frameworks. I think we'd have more value from these scripts if we can have self contained .sh files that we can run from the place where we want to enable evpn and copy them around (in containers or pods). This would also serve as docs to the users.

e2etests/tests/testdata/evpn-setup.sh
```diff
@@ -0,0 +1,530 @@
...
+# VRF-VNI associations from zebra before route-targets can be configured.
+#
+# Both sides use the same route-target values (EVPN_FRR_K8S_ASN:VNI) so that
+# routes are imported/exported correctly regardless of eBGP/iBGP.
+#
+# NOTE: This configures FRR via vtysh (runtime config). If the test framework
+# later calls UpdateBGPConfigFile, these changes will be overwritten. For
+# integration tests (Phase 7), EVPN config should be included in the BGP
+# config template instead.
+configure_external_frr() {
```
fedepaol: I'd avoid this and go putting the configurations from the tests. We are currently relying on metallb's infra, but I think it should be doable to change that to accept a raw frr config bit that we can then setup from the tests.

README.md
```diff
@@ -202,6 +202,69 @@ spec:
...
+      evpn:
+        l3vni:
+          vni: 2000
+          rd: "64512:2000"
+          importRTs: ["64512:2000"]
+          exportRTs: ["64512:2000"]
+          advertisePrefixes: ["unicast"]
+```
+
+**Note:** EVPN requires host networking setup (VXLAN interfaces, bridges, VRFs) outside of frr-k8s. See `hack/evpn-setup.sh` for a reference configuration.
```
fedepaol: see my other comments on the testing scripts. If we extract the host configuration logic, we can refer directly to that script.

fedepaol (commented):
Left comments on the testing infra (still need to go on the tests implementation which I think can be affected by the current review)

api/v1beta1/frrconfiguration_types.go
jcaamano: https://kubernetes.io/docs/reference/using-api/cel/#resource-constraints

TLDR

Evaluating CEL has a cost for API server and there are some compile time rules to budget it disallowing excessive cost. For example, cost of CEL rules over unbounded list is infinite and disallowed.

fedepaol: ah, TIL! Thanks for explaning

e2etests/tests/evpn.go
jcaamano: The file is already `evpn.go` so just renamed the helper functions to remove the `evpn` prefix.

e2etests/tests/testdata/evpn-setup.sh
fedepaol: I probably want's clear, but I'd move this logic to the Go side as all the other bits that configure frr

e2etests/tests/testdata/evpn-setup.sh
```diff
@@ -0,0 +1,488 @@
+#!/bin/bash
```
fedepaol: I would translate this to Go. After sending the script to the pods / external containers, we can use the executor package to run the script from there. I'd keep only `evpn-node-setup.sh` in bash

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,756 @@
...
+	if result.VNI != vni {
+		return fmt.Errorf("VNI %d not found in EVPN state, got %d", vni, result.VNI)
+	}
+	return nil
+}
+
+func nodeL2VNIMacs(nodes []corev1.Node) ([]string, error) {
+	macs := make([]string, 0, len(nodes))
+	for _, n := range nodes {
+		out, err := executor.Host.Exec(executor.ContainerRuntime, "exec", n.Name,
```
fedepaol: we should use executor.ForContainer here with the name of the node or even better ForPod with frrk8s pods as they are host networked. This will make teh tests more portable

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,756 @@
...
+					return nil
+				}
+			}
+			return fmt.Errorf("neighbor %s does not have l2vpnEvpn address family, has: %v", neighborIP, n.AddressFamilies)
+		}
+	}
+	return fmt.Errorf("neighbor %s not found", neighborIP)
+}
+
+func vniExists(exec executor.Executor, vni uint32) error {
```
fedepaol: might make sense to make an frr package under e2etests/pkg for this

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,756 @@
...
+		mac := strings.TrimSpace(out)
+		if mac == "" {
+			return nil, fmt.Errorf("empty MAC for node %s", n.Name)
+		}
+		macs = append(macs, mac)
+	}
+	return macs, nil
+}
+
+func checkType2Routes(exec executor.Executor, expectedMACs []string) error {
```
fedepaol: might make sense to make an frr package under e2etests/pkg for this

Also, this is more existType2RoutesForMacs

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,756 @@
...
+			missing = append(missing, mac)
+		}
+	}
+	if len(missing) > 0 {
+		return fmt.Errorf("type-2 routes missing for MACs: %v", missing)
+	}
+	return nil
+}
+
+func checkType5Routes(exec executor.Executor, expectedRoutes map[string]string) error {
```
fedepaol: same as type2 routes

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,756 @@
...
+		}
+		if !found {
+			return fmt.Errorf("type-5 route %s expected next-hop %s, got %v", routeKey, expectedNH, nhs)
+		}
+	}
+	return nil
+}
+
+func ping(container, targetIP string) error {
+	out, err := executor.Host.Exec(executor.ContainerRuntime, "exec", container,
```
fedepaol: same considerations as above, better use executor.ForContainer or ForPod depending on the target

e2etests/pkg/frr/evpn.go
```diff
@@ -0,0 +1,181 @@
...
+	evpnConfig, err := renderEVPNConfig(data)
+	if err != nil {
+		return fmt.Errorf("rendering EVPN config: %w", err)
+	}
+
+	if err := c.UpdateBGPConfigFile(baseConfig + evpnConfig); err != nil {
+		return fmt.Errorf("writing BGP config: %w", err)
+	}
+
+	// The VRF/VNI association is a zebra command that cannot go through
```
fedepaol: I didn't realize this 😅 I am going to change the metallb e2e tests package to always write the frr.conf, which is more adherent to what we do in metallb / frrk8s anyways

e2etests/pkg/frr/evpn.go
```diff
@@ -0,0 +1,181 @@
+// SPDX-License-Identifier:Apache-2.0
```
fedepaol, line 1: let's name this evpn.go

e2etests/pkg/infra/evpn_setup.go
```diff
@@ -0,0 +1,283 @@
...
+
+// EVPNContainer is set during SetupEVPN to the ebgp-single-hop container used
+// for EVPN peering.
+var EVPNContainer *frrcontainer.FRR
+
+// SetupEVPN configures EVPN Linux networking (bridge, VXLAN, VNI, VRF, access
+// ports) on all cluster nodes and the ebgp-single-hop container. It returns an
+// EVPN that provides name-based address lookups for each target.
+func SetupEVPN(cs clientset.Interface, cfg EVPNConfig) (*EVPN, error) {
+	ebgpSingleHop := findContainer("ebgp-single-hop")
```
fedepaol: let's make the name of the container a parameter, it might be useful to change it from test execution (on d/s for example, where we only run one container at the time)

fedepaol: also, let's move this to a separate function (EVPNPeer() or something like that ), no need to have it as a global variable. In the test we can fetch in in beforeeach

e2etests/pkg/infra/evpn_setup.go
```diff
@@ -0,0 +1,283 @@
...
+	}
+
+	extIndex := len(nodes) + 1
+	extEnv := buildTargetEnv(cfg, externalFRR.Ipv4, extIndex)
+	extExec := executor.ForContainer(externalFRR.Name)
+	if err := runTargetSetup(extExec, extEnv); err != nil {
+		return fmt.Errorf("node setup on external FRR %s: %w", externalFRR.Name, err)
+	}
+
+	time.Sleep(5 * time.Second)
```
fedepaol: can we wait for something instead of sleep-ing?

e2etests/pkg/infra/evpn_setup.go
```diff
@@ -0,0 +1,283 @@
...
+		env := buildCleanupEnv(cfg, nodeIP, i+1)
+		if err := runTargetSetup(podExec, env); err != nil {
+			return fmt.Errorf("cleanup on %s: %w", node.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func podForNode(pods []*corev1.Pod, nodeName string) (*corev1.Pod, error) {
```
fedepaol: nit: move this under the k8s package

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,552 @@
...
+	L3IPv6PrefixFmt: "fc00:200:%d::1/64",
+	Bridge:          evpnBridge,
+	Vxlan:           evpnVxlan,
+}
+
+var _ = ginkgo.DescribeTableSubtree("EVPN",
+	func(peerFamily ipfamily.Family, advertiseFamily ipfamily.Family) {
+		var cs clientset.Interface
+
+		evpnL2Config := evpnL2Config
```
fedepaol: can we use a different name (maybe *Template ) for the global ones?

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,552 @@
...
+			ginkgo.GinkgoHelper()
+			evpnCfg = cfg
+
+			ginkgo.By("Setting up EVPN networking infrastructure")
+			var err error
+			evpnInfra, err = infra.SetupEVPN(cs, cfg)
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("Pairing external FRR with nodes (EVPN)")
+			advertiseFamily := ipfamily.DualStack
```
fedepaol: isn't this driven by advertiseFamily above? Like, we set cfg.L3IPv6PrefixFmt / cfg.L3PrefixFmt based on the global advertiseFamily, so I would use that instead of this indirection. Also, let's not shadow variable names

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,552 @@
...
+						"EVPN address family not active on pod %s for peer %s", pod.Name, peerIP)
+				}
+			}
+
+			ginkgo.By("Validating EVPN type-5 routes are received on external FRR")
+			expectedRoutes := map[string]string{}
+			for _, node := range nodes {
+				nodeIP, err := k8s.NodeIPForFamily(node, ipfamily.IPv4)
+				Expect(err).NotTo(HaveOccurred())
+				if evpnCfg.L3PrefixFmt != "" {
```
fedepaol: same here, we could use ipfamily

e2etests/tests/evpn.go
```diff
@@ -0,0 +1,552 @@
...
+				for _, node := range nodes {
+					l3IP := evpnInfra.L3PrefixIP(node.Name)
+					Eventually(func() error {
+						return pingVRF(infra.EVPNContainer, evpnL3VRF, l3IP)
+					}, 30*time.Second, time.Second).ShouldNot(HaveOccurred(),
+						"L3 ping from external FRR to node %s (%s) via VRF %s failed", node.Name, l3IP, evpnL3VRF)
+				}
+			}
+
+			if evpnCfg.L3IPv6PrefixFmt != "" {
```
fedepaol: this is quite similar to the previous block, with the difference of how we get the ip, I'd refactor them

e2etests/pkg/frr/evpn.go
fedepaol: with this in https://github.com/metallb/metallb/pull/3046 if you consume the latest version of e2es you can get rid of the vtysh addon and generate the config as a whole

e2etests/pkg/infra/evpn_setup.go
jcaamano: changed to `ibgp-single-hop` driven by the test and not this setup harness.

e2etests/pkg/infra/evpn_setup.go
jcaamano: not needed, removed

e2etests/tests/evpn.go
jcaamano: `PairWithNodesForEVPN` is not concerned with addresses at all so why pass that.

I will pass the global `advertiseFamily` if you are concerned with the indirection.

e2etests/pkg/frr/evpn.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return fmt.Errorf("rendering EVPN config: %w", err)
+	}
+
+	vrfPreamble, err := renderVRFPreamble(data.L3VNI)
+	if err != nil {
+		return fmt.Errorf("rendering VRF preamble: %w", err)
+	}
+
+	fullConfig := insertBeforeRouterBGP(baseConfig, vrfPreamble) + evpnConfig
```
fedepaol, line 85: now that we are rendering the config as a whole, we can use the template instead of appending I think?
Or, what's the reason for appending / adding after?

e2etests/pkg/infra/evpn_setup.go
```diff
@@ -0,0 +1,293 @@
...
+	extExec := executor.ForContainer(externalFRR.Name)
+	if err := runTargetSetup(extExec, extEnv); err != nil {
+		return fmt.Errorf("node setup on external FRR %s: %w", externalFRR.Name, err)
+	}
+
+	return nil
+}
+
+func cleanupEVPNNetworking(cs clientset.Interface, nodes []corev1.Node, externalFRR *frrcontainer.FRR, cfg EVPNConfig) error {
+	if externalFRR != nil {
```
fedepaol, line 215: I don't think this can be nil

e2etests/pkg/frr/evpn.go
jcaamano: We are using upstream `BGPPeersForAllNodes` to generate the unicast config. We are using our own templates here `evpnConfigTemplate`/`evpnVRFTemplate` to generate the evpn config. We then append both configs. The alternative here is having a unique full template here, copying, not reusing, what `BGPPeersForAllNodes` already does. Not sure if that is what you are suggesting.

However, when appending, we need to insert VRF-VNI block before the any router block, so what we actually do is insert the `evpnVRFTemplate` and then append the `evpnConfigTemplate`. Reson why explain in the `insertBeforeRouterBGP` doctext:
```
// insertBeforeRouterBGP inserts preamble text before the first "router bgp"
// line in config. This is needed because vtysh --mark (used by frr-reload.py)
// cannot parse VRF definitions that appear after router bgp blocks.
func insertBeforeRouterBGP(config, preamble string) string {
```

I though using `frr.conf` for this instead of `bgp.conf` would help but it didn't: VRF blocks after router blocks fail to apply.

e2etests/pkg/infra/evpn_setup.go
jcaamano: If this fails in `BeforeEach`
```
evpnFRRContainer = infra.FindContainer(evpnFRRContainerName)
			Expect(evpnFRRContainer).NotTo(BeNil(), "%s container not found in FRRContainers", evpnFRRContainerName)
```
AfterEach is executed anyway
```
			err := infra.CleanupEVPN(cs, evpnFRRContainer, evpnCfg)
			Expect(err).NotTo(HaveOccurred())
```

This nil check makes `CleanupEVPN` idempotent in this case. If you prefer I can remove the nil check and use an extra flag avoid calling `CleanupEVPN`.

e2etests/pkg/frr/evpn.go
fedepaol: right, I missed the `BGPPeersForAllNodes` bit. I guess we can live with it :-)

fedepaol:
thanks for the replies (and for the patience!), please clean up the commit history and I'll merge

jcaamano:
squashed and rebased, no conflicts

fedepaol:
Thanks!

PR #432, opened 2026-04-27
CI: add a daily ci run

karampok:
Having a periodic run is needed because:
1. in our CI we fetch the metallb main branch, so if something breaks it will only be visible when a new PR lands in the repo.
2. flakes will be found.

Revives #325.

/kind cleanup

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -13,6 +13,8 @@ on:
       - "main"
       - "v**"
   workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * *'
 
 concurrency:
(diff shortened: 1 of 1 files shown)


fedepaol:
@karampok I feel like there's something odd with the lanes invoked here.
I guess the 3 we see failing are just the old ones being renamed (and that's not a big deal, we'll adjust the list of mandatory jobs), but many are missing / not being called at all

karampok:
@fedepaol  now  the implementation should be ok and simple.
 (bit strange why it was not so simple but long time ago to remember)
Thanks

karampok:
@fedepaol @oribon  this is getting again stale work.

PR #445, opened 2026-05-15
Generate fallback BGP router-id on IPv6-only nodes

cgoncalves:

/kind bug

On IPv6 single-stack clusters where nodes have no IPv4 addresses, FRR defaults the BGP router-id to 0.0.0.0. This violates RFC 6286 and RFC 4271, causing remote peers to reject the session with "Bad BGP Identifier".

Derive a deterministic router-id from the node hostname using CRC32 when no explicit router-id is configured, matching the approach already used by MetalLB's native BGP implementation.

**Release note**:

```release-note
When no explicit BGP router-id is configured, frr-k8s now derives a deterministic router-id from the node hostname, fixing BGP session establishment failures on IPv6 single-stack clusters where FRR would default to the invalid 0.0.0.0 identifier.
```

### DIFF
--- a/internal/frr/frr.go
+++ b/internal/frr/frr.go
@@ -4,7 +4,10 @@ package frr
 
 import (
 	"context"
+	"encoding/binary"
 	"fmt"
+	"hash/crc32"
+	"net"
 	"os"
 	"strings"
 	"sync"
@@ -31,9 +34,10 @@ type Status struct {
 }
 
 type FRR struct {
-	reloadConfig    chan reloadEvent
-	Status          Status
-	onStatusChanged StatusChanged
+	reloadConfig     chan reloadEvent
+	Status           Status
+	onStatusChanged  StatusChanged
+	fallbackRouterID string
 	sync.Mutex
 }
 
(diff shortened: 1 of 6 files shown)


internal/frr/frr.go
```diff
@@ -51,10 +55,21 @@ func (f *FRR) ApplyConfig(config *Config) error {
 
 	// TODO add internal wrapper
 	config.Hostname = hostname
+	for _, r := range config.Routers {
+		if r.RouterID == "" {
+			r.RouterID = hashRouterID(hostname)
+		}
+	}
```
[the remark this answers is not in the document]
cgoncalves: Fixed. Recheck.

internal/frr/frr.go
```diff
@@ -51,10 +55,21 @@ func (f *FRR) ApplyConfig(config *Config) error {
...
 	f.reloadConfig <- reloadEvent{config: config}
 	return nil
 }
 
+func hashRouterID(hostname string) string {
+	buf := new(bytes.Buffer)
+	binary.Write(buf, binary.LittleEndian, crc32.ChecksumIEEE([]byte(hostname)))
+	return net.IP(buf.Bytes()).String()
+}
```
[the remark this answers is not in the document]
cgoncalves: Fixed. Recheck.

internal/frr/frr.go
```diff
@@ -51,10 +54,22 @@ func (f *FRR) ApplyConfig(config *Config) error {
 
 	// TODO add internal wrapper
 	config.Hostname = hostname
+	routerID := hashRouterID(hostname)
+	for _, r := range config.Routers {
+		if r.RouterID == "" {
```
fedepaol: the problem here is that this is taking into account the absence of user provided routerid (as in, coming from the api) while we should probably implement this only when frr is not able to calculate it (from what we saw, on ipv6 only nodes.
Always overriding it seems too big for a change

cgoncalves: Fair point. Even in the case where the node has IPv4 addresses but none is in the interface used for BGP peering, FRR's router-id selection is global. It picks the highest IPv4 address across all interfaces on the node, regardless of which interface is used for BGP peering.

internal/frr/frr.go
```diff
@@ -51,10 +54,46 @@ func (f *FRR) ApplyConfig(config *Config) error {
...
+			}
+		}
+	}
 	f.reloadConfig <- reloadEvent{config: config}
 	return nil
 }
 
+var netInterfaces = net.Interfaces
+
+func hasIPv4Address() bool {
```
fedepaol, line 71: where we iterate over all the network interfaces every time we process the config, which is less than optimal. Imo we can calculate the routerid (if needed) when the process starts (or the frr object is created, I don't remember if we have one)

internal/frr/frr.go
```diff
@@ -51,10 +54,46 @@ func (f *FRR) ApplyConfig(config *Config) error {
 
 	// TODO add internal wrapper
 	config.Hostname = hostname
+	// On IPv6-only nodes, FRR defaults router-id to 0.0.0.0 (RFC 6286 violation).
+	if !hasIPv4Address() {
+		routerID := hashRouterID(hostname)
```
fedepaol: we should produce a log when this happen (assuming we'll move it somewhere else that is triggered only once).

internal/frr/frr.go
cgoncalves: The `osHostname()` call can return an `error` but in practice it never fails (it's a simple syscall). It is used in `ApplyConfig()` without catching for errors. Catching errors is always good practice of course. I'll need to change the signature of `NewFRR` so that the error is propagated, plus tests. A lot of churn for a failure that realistically won't happen, but so be it.

internal/frr/frr.go
```diff
@@ -51,21 +55,60 @@ func (f *FRR) ApplyConfig(config *Config) error {
...
+func NewFRR(ctx context.Context, onStatusChanged StatusChanged, debounceTimeout time.Duration) (*FRR, error) {
+	l := logging.GetLogger()
 	res := &FRR{
 		reloadConfig:    make(chan reloadEvent),
 		onStatusChanged: onStatusChanged,
 	}
 
+	// On IPv6-only nodes, FRR defaults router-id to 0.0.0.0 (RFC 6286 violation).
+	if !hasIPv4Address() {
+		hostname, err := osHostname()
```
fedepaol: I would embed this in hashRouterID

cgoncalves:
None of these failures are related to the PR changes: flaky tests and vulnerability in net/http2.

fedepaol:
LGTM, thanks!

PR #446, opened 2026-05-19
Add next-hop support to FRRConfiguration

trozet:
Allow FRRConfiguration users to set an advertised BGP next-hop through the typed toAdvertise API. The new field carries separate IPv4 and IPv6 addresses and converts them into route-map next-hop statements during FRR rendering.

Reject invalid next-hop addresses and cases where the requested next-hop family is not enabled for the neighbor. When multiple FRRConfigurations merge into the same neighbor, keep matching values and fail on conflicting next-hop settings.

Regenerate deepcopy methods, CRDs, chart CRDs, all-in-one manifests, and API docs, and add conversion, merge, and render coverage.

/kind feature

Allows specifying next hop through FRR-K8S API. This is needed when the next hop cannot be inferred as the FRR speaker address. For example, in OVN-Kubernetes DPU deployments, FRR may run on the DPU, while the advertised routes should use the Host shared gateay address as the next hop.

**Release note**:

```
FRRConfiguration now supports setting IPv4 and IPv6 BGP next-hop addresses for advertised prefixes via `toAdvertise.nextHop`.
```


internal/controller/merge_test.go
```diff
@@ -1022,6 +1022,72 @@ func TestMergeNeighbors(t *testing.T) {
...
+					Name:     "65040@192.0.1.20",
+					ASN:      "65040",
+					Addr:     "192.0.1.20",
+					Outgoing: frr.AllowedOut{
+						PrefixesV4: []string{"192.0.2.0/24"},
+						NextHopV4:  "192.0.2.2",
+					},
+				},
+			},
+			err: fmt.Errorf("multiple ipv4 next hops specified"),
```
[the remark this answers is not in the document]
trozet: merge_test.go only checks whether an error exists, not the exact message. But will fix the message anyway.

trozet:
@fedepaol @oribon do you guys mind taking a look?

internal/controller/merge_test.go
```diff
@@ -1022,6 +1022,72 @@ func TestMergeNeighbors(t *testing.T) {
...
+			name: "Multiple next hops for a prefix family",
+			curr: []*frr.NeighborConfig{
+				{
+					IPFamily: ipfamily.IPv4,
+					Name:     "65040@192.0.1.20",
+					ASN:      "65040",
+					Addr:     "192.0.1.20",
+					Outgoing: frr.AllowedOut{
+						PrefixesV4: []string{"192.0.2.0/24"},
+						NextHopV4:  "192.0.2.1",
```
karampok, line 1035: can we extend with some ipv6 entries the unit test?

karampok (commented):
Is it possible to have e2e for that?

trozet:
@karampok is CI stuck?

internal/controller/api_to_config_test.go
```diff
@@ -351,6 +351,9 @@ func TestConversion(t *testing.T) {
 													Prefixes: []string{"192.0.2.0/24"},
 													Mode:     v1beta1.AllowRestricted,
 												},
+												NextHop: v1beta1.NextHop{
```
oribon, line 354: this only tests the happy path, can we add the cases for: ipv6/dualstack/errors?

internal/controller/merge.go
```diff
@@ -124,6 +133,16 @@ func mergeAllowedOut(r, toMerge frr.AllowedOut) (frr.AllowedOut, error) {
 	return res, nil
 }
 
+func mergeNextHop(curr, toMerge, ipFamily string) (string, error) {
```
oribon: can we omit the ipFamily from here, and have the caller wrap the err with the family?

e2etests/pkg/routes/routes.go
```diff
@@ -53,6 +53,36 @@ func CheckNeighborHasPrefix(neighbor frrcontainer.FRR, vrf, prefix string, nodes
...
+	}
+
+	expected := net.ParseIP(nextHop)
+	if expected == nil {
+		return fmt.Errorf("invalid next hop %q", nextHop)
+	}
+	for _, actual := range route.NextHops {
+		if actual.Equal(expected) {
+			return nil
+		}
```
oribon: from the api doc "IPv4 is **the** next-hop address" I'd expect it to be the only next hop for the route, if that's the case should we enforce it here (making sure route.NextHops is len=1)?

e2etests/tests/validate.go
```diff
@@ -61,6 +63,26 @@ func ValidatePrefixesForNeighborVRF(neigh frrcontainer.FRR, nodes []v1.Node, vrf
...
 }
 
+func ValidatePrefixesForNeighborWithNextHop(neigh frrcontainer.FRR, nodes []v1.Node, nextHop string, prefixes ...string) {
+	ginkgo.By(fmt.Sprintf("checking prefixes %v for %s with next hop %s", prefixes, neigh.Name, nextHop))
+	nextHopIP := net.ParseIP(nextHop)
+	Expect(nextHopIP).NotTo(BeNil(), "invalid next hop %q", nextHop)
+
+	defaultNextHops, err := e2ek8s.NodeIPsForFamily(nodes, ipfamily.ForAddress(nextHopIP), neigh.RouterConfig.VRF)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(defaultNextHops).NotTo(ContainElement(nextHop), "configured next hop must differ from the default node next hops")
```
oribon: what's the reason for having this here, and should we be enforcing this at the code level? I think this is the only place this restriction shows up

e2etests/tests/split.go
oribon, line 1: can you also add a couple of cases to webhooks.go?

internal/frr/frr_test.go
```diff
@@ -122,6 +122,7 @@ func TestSingleSession(t *testing.T) {
 								"192.169.1.0/24",
 								"192.170.1.0/22",
 							},
+							NextHopV4: "192.168.1.1",
```
oribon, line 172: can we have this on a separate case (+ cover ipv6)?

oribon (commented):
thanks! a couple of small nits/questions, code looks good (and sorry for not responding, was on vacation)

e2etests/tests/validate.go
trozet: It shouldn't be there. Removed.

oribon (commented):
looks good, can you rebase and resolve the ci fails?

oribon (approved):
lgtm, thanks!

trozet:
thanks @oribon !!

PR #451, opened 2026-06-03
Add linux/riscv64 build

mmoll:
This works under Ubuntu 24.04, I hope the build under Ubuntu 22.04 (used in the pipeline currently) will get through.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -88,7 +88,7 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
           file: Dockerfile
-          platforms: linux/amd64,linux/arm64,linux/s390x,linux/ppc64le,linux/arm/v7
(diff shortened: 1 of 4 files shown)


oribon:
the changes here make sense, but have you tried this locally?
the reason I'm asking is because it seems frr itself https://quay.io/repository/frrouting/frr?tab=tags doesn't have a riscv build, so I wonder if I'm missing something 😅

mmoll:
opened https://github.com/FRRouting/frr/pull/22256 and will revisit this PR then... 😮‍💨

oribon:
@mmoll thanks a lot!

oribon:
hey @mmoll, I see that https://github.com/FRRouting/frr/pull/22256 merged, do we want to proceed with this pr?

mmoll:
@oribon there is not tagged build with this yet and I also think only frr >= 10.6 will get riscv64 builds, so I guess this would need an update here first?

oribon:
ah yes, so let's wait for when we bump it here first. thanks!

mmoll:
waits for #486

PR #461, opened 2026-06-24
Added AsPathPrepend parameter

psasnal:
/kind feature

This PR introduces support for prefix-level AS-Path prepending within the `FRRConfiguration` CRD to enable granular inbound traffic engineering.

It adds the `prefixesWithAsPathPrepend` configuration under `ToAdvertise` for BGP neighbors, allowing users to target specific prefixes and define their desired AS-Path prepend counts. The controller translates these definitions into FRR route-maps, dynamically generating `set as-path prepend` statements matching the configured prefixes.

Relates to metallb/metallb#808

Updated implementation from the initial neighbor-level approach to prefix-level prepending based on review feedback.

**Release note**:
```release-note
ADD: Support for prefix-level AS-Path prepending configuration on BGP neighbors to enable inbound traffic engineering.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for AS-path prepending on selected advertised CIDR prefixes.
  * Configure prepend counts from 1–10 using `FRRConfiguration`.
  * Supports router or neighbor local ASN selection and IPv4/IPv6 advertisements.
  * Added validation for invalid, duplicate, or unauthorized prefix configurations.

* **Bug Fixes**
  * Conflicting AS-path prepend settings are now rejected during configuration merging.

* **Tests**
  * Added coverage for eBGP advertisement behavior, configuration generation, validation, and route-policy merging.


oribon (commented):
I haven't gone over the code yet, but I have a question regarding the API - I don't know much about AS path prepending, but why did we go with putting it on the Neighbor scope vs a per prefix scope (e.g like localprefs / communities)? from my minimal understanding, the latter seems more aligned with what prepending does (and the existing API), what were your considerations (and wdyt)?

psasnal:
@oribon 
That's a completely valid point, especially since things like communities or localpref live at the advertisement/prefix level. However, going with the neighbor scope here felt like a better fit for a couple of practical reasons:
- Link-Level Traffic Engineering: The main use case here is setting up active/backup paths. If an operator is multi-homed to two upstream switches, they usually want to deprioritize the entire backup link. It’s way cleaner to say "prepend everything going to this neighbor" rather than forcing them to micromanage that config on every single prefix advertisement.

- Architectural Fit: We already handle other path-influencing session overrides at the neighbor scope, like LocalASN masquerading. Keeping AsPathPrepend at the neighbor level follows that exact same architectural pattern. Plus, because frr-k8s generates a dedicated outbound route-map per peer ({{neighbor.ID}}-out), this struct is the natural place for it—especially since the template already needs to check the neighbor's LocalASN state to decide which ASN to actually prepend.

Those were the main design considerations behind going with the neighbor scope for this feature.

psasnal:
This PR failed govulncheck because of golang.org/x/net@0.39.0. I upgraded it to v0.56.0 to resolve govulncheck CVEs as my most recent commit here.

oribon:
> @oribon That's a completely valid point, especially since things like communities or localpref live at the advertisement/prefix level. However, going with the neighbor scope here felt like a better fit for a couple of practical reasons:
> 
> * Link-Level Traffic Engineering: The main use case here is setting up active/backup paths. If an operator is multi-homed to two upstream switches, they usually want to deprioritize the entire backup link. It’s way cleaner to say "prepend everything going to this neighbor" rather than forcing them to micromanage that config on every single prefix advertisement.

users already "micromanage" this for local prefs and communities, so I'm not sure how strong this holds. **are there any other use cases we would be missing by doing the opinionated choice of enabling only prepending on the neighbor level?**

> * Architectural Fit: We already handle other path-influencing session overrides at the neighbor scope, like LocalASN masquerading. Keeping AsPathPrepend at the neighbor level follows that exact same architectural pattern. Plus, because frr-k8s generates a dedicated outbound route-map per peer ({{neighbor.ID}}-out), this struct is the natural place for it—especially since the template already needs to check the neighbor's LocalASN state to decide which ASN to actually prepend.
> 

we shouldn't constrain API contracts on implementation details. even if we go that route, local asn is a "neighbor" command in FRR, whereas this is a "route map set" (which favors doing per prefix and not neighbor). aren't community and localpref "path-influencing" as well?

wdyt?

psasnal:
I looked at the distinction between a native neighbor command like local-as and a route-map set action - tying the API contract to our current template implementation definitely isn't the right move long-term.

I think you are also right on the use case side. Locking this to the neighbor level would completely cut off asymmetric traffic engineering, like prepending only a specific prefix on a session while advertising others normally. Since users are already managing things like communities and localprefs at the advertisement level, keeping the API consistent makes a lot more sense.

I'm down to pivot and move AsPathPrepend over to the advertisement/prefix scope. Before I start tearing into the templates and merge logic, do you have a specific schema layout in mind for where you'd like this to sit?

fedepaol:
@psasnal sorry if I jump in, but I couldn't avoid noticing that the previous comment https://github.com/metallb/frr-k8s/pull/461#issuecomment-4834885112 seems heavily AI generated.

While contributions to the project are more than appreciated, we (as maintainers) would appreciate contributors doing the extra step of prefiltering the content. Short-cutting us to the answers generated by a machine doesn't add any value in terms of contribution and makes no difference in us reviewing claude's output without the need of a contributor.

I am writing it here just to prevent this kind of interaction will leak to the code review phase, so the ask is not to use AI to get help, but to filter it's output with a critic eye and re-elaborate it before submitting to us. If you did already, please forgive this long message.

psasnal:
@fedepaol
I appreciate the heads-up, but the technical decisions and responses here are entirely driven by me and my team. We actually debated your feedback internally before I posted that response, as pivoting to the prefix scope is a major shift from our original design. I simply take the time to format my thoughts cleanly to ensure clear communication.

@oribon
Your point about the API contract and route-map actions makes complete sense. I went ahead and pivoted the code. I added `withAsPathPrepend` right alongside `withLocalPref` and `withCommunity` to keep the schema consistent, updated the template logic, and re-ran the code generators. Let me know what you think of the new layout.

fedepaol:
> @fedepaol I appreciate the heads-up, but the technical decisions and responses here are entirely driven by me and my team. We actually debated your feedback internally before I posted that response, as pivoting to the prefix scope is a major shift from our original design. I simply take the time to format my thoughts cleanly to ensure clear communication.
> 

Fair enough then, all good! I wanted to ensure everything was going on the right direction. Thanks again for the help!

api/v1beta1/frrconfiguration_types.go
```diff
@@ -338,6 +344,22 @@ type CommunityPrefixes struct {
...
+	// +kubebuilder:validation:Format="cidr"
+	Prefixes []string `json:"prefixes,omitempty"`
+
+	// AsPathPrepend dictates how many additional times the local AS number
+	// is prepended to the AS-path on outgoing BGP updates. This is used to
+	// artificially lengthen the AS-path so external peers route traffic to
+	// other paths (e.g., creating active/standby routing).
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=10
+	AsPathPrepend uint8 `json:"asPathPrepend,omitempty"`
```
oribon, line 374: we need to add a runtime check in prefixesWithAsPathPrependToFRR: if prefixes.AsPathPrepend == 0 { return error } so that this is never 0

internal/controller/api_to_config.go
```diff
@@ -429,6 +439,56 @@ func prefixesWithCommunityToFRR(toAdd map[string]frr.CommunityPrefixList, neighb
...
+	for _, prefixes := range toAdvertise.PrefixesWithAsPathPrepend {
+		var asnToRepeat string
+		if neighbor.LocalASN > 0 {
+			asnToRepeat = fmt.Sprintf("%d", neighbor.LocalASN)
+		} else {
+			asnToRepeat = fmt.Sprintf("%d", routerASN)
+		}
+
+		count := int(prefixes.AsPathPrepend)
+		AsPathPrependASNs := make([]string, count)
```
oribon: nit: should be asPathPrependASNs (start with lowercase)

internal/controller/merge_test.go
```diff
@@ -1762,6 +1762,17 @@ func localPrefPrefixListFor(neigID string, localPref int, ipFamily string, prefi
 	}
 }
 
+func asPathPrependPrefixListFor(neigID string, asPathPrepend []string, ipFamily string, prefixes []string) frr.AsPathPrependPrefixList {
```
oribon: no test cases exercise mergeAsPathPrependPrefixLists. The merge path handles combining AS path prepend configs from multiple FRRConfiguration CRs targeting the same neighbor — worth a test case (same prepend count merging prefixes, different counts staying separate).

api/v1beta1/frrconfiguration_types.go
psasnal: fixed!

internal/controller/merge_test.go
psasnal: Good feedback, actually I wrote unit tests and figured out I need to correct the implementation as well :) I also had to make some small changes and fixes in merge_test.go file.

internal/controller/api_to_config.go
```diff
@@ -429,6 +439,60 @@ func prefixesWithCommunityToFRR(toAdd map[string]frr.CommunityPrefixList, neighb
...
+		if asnPrependCount == 0 {
+			return nil, fmt.Errorf("AsPathPrepend can't be zero")
+		}
+
+		var asnToRepeat string
+		if neighbor.LocalASN > 0 {
+			asnToRepeat = fmt.Sprintf("%d", neighbor.LocalASN)
+		} else {
+			asnToRepeat = fmt.Sprintf("%d", routerASN)
+		}
```
oribon: nit: avoid the else by doing:
```go
asnToRepeat := fmt.Sprintf("%d", routerASN)
if neighbor.LocalASN > 0 {
    asnToRepeat = fmt.Sprintf("%d", neighbor.LocalASN)
}

File: internal/controller/api_to_config.go [hunk 5]
```diff
@@ -429,6 +439,60 @@ func prefixesWithCommunityToFRR(toAdd map[string]frr.CommunityPrefixList, neighb
...
+
+		asPathPrependASNs := make([]string, asnPrependCount)
+		for i := 0; i < asnPrependCount; i++ {
+			asPathPrependASNs[i] = asnToRepeat
+		}
```
oribon: I think it would be cleaner having this logic in frr/config.go , then this file knows less about frr internal representation and it would simplify both the merge logic and this function, what I mean is:
```go
type AsPathPrependPrefixList struct {
    PrefixList
    PrependASN   uint32
    PrependCount uint8
}
```
probably suffices with:
```go
func (pl AsPathPrependPrefixList) SetStatement() string {
    asn := fmt.Sprintf("%d", pl.PrependASN)
    parts := make([]string, pl.PrependCount)
    for i := range parts {
        parts[i] = asn
    }
    return fmt.Sprintf("set as-path prepend %s", strings.Join(parts, " "))
}
```

internal/controller/api_to_config.go
oribon, line 453: also, is this the only thing we can do (i.e does frr deny it otherwise) or this is our opinionated choice? does it make sense to document this as part of the api?

e2etests/go.sum
oribon, line 1: can we add some e2es? I thought we had a couple on a previous iteration 😅 ?

internal/controller/api_to_config.go
psasnal: Good point. FRR does technically let you prepend whatever ASNs you want (like `set as-path prepend 123 456`), but doing that in the real world is considered BGP spoofing or a misconfiguration. If you inject foreign ASNs, it usually triggers loop-prevention drops in upstream peers.

By designing the API to just take a multiplier (e.g., 3) and strictly repeating the local ASN, we are making an opinionated but very safe choice. This allows users to lengthen the AS-path for traffic engineering while preventing accidental routing issues upstream.

I will update the API docs on the struct to explicitly call out this behavior so it's clear to users.

internal/controller/api_to_config.go
psasnal: ok

e2etests/go.sum
psasnal: I'm on it! I wanted to get the core logic and unit tests dialed in first. I'll work on adding the e2e tests next and will ping you when they are pushed.

internal/controller/api_to_config.go
oribon: cool that's another point, but I meant the part where we choose localASN over routerASN if localASN > 0 (sorry I wasn't clear in the previous comment)

psasnal: FRR's route-maps will technically let you prepend the primary ASN even if `local-as` is configured on the neighbor. So in a way, it is an opinionated choice.
However, I decided to automatically use LocalASN (when set) as a guardrail. Since the main use case for local-as is masquerading, if the controller blindly prepended the primary ASN, it would leak the hidden ASN into the path and likely cause the peer to drop the routes. Automatically falling back to the LocalASN safely maintains the masquerade. I will update the API docs to explicitly call out this behavior so it's transparent to the user.

e2etests/tests/validate.go
```diff
@@ -175,6 +176,69 @@ func ValidateNeighborLocalPrefForPrefix(neigh frrcontainer.FRR, prefix string, e
 	}, 5*time.Second, time.Second).ShouldNot(HaveOccurred())
 }
 
+func ValidateNeighborAsPathPrependForPrefix(neigh frrcontainer.FRR, prefix string, expectedAsPathPrepend uint8, ipfam ipfamily.Family) {
+	if !strings.Contains(neigh.Name, "ebgp") {
+		return // asPathPrepend is valid only for iBGP connections
```
oribon: nit: for eBGP

e2etests/tests/validate.go
```diff
@@ -175,6 +176,69 @@ func ValidateNeighborLocalPrefForPrefix(neigh frrcontainer.FRR, prefix string, e
...
+
+		if asPathPrepend != expectedAsPathPrepend {
+			return fmt.Errorf("asPathPrepend %d for prefix %s on neighbor %s does not equal %d", asPathPrepend, prefix, neigh.Name, expectedAsPathPrepend)
+		}
+
+		return nil
+	}, 5*time.Second, time.Second).ShouldNot(HaveOccurred())
+}
+
+// TODO move to metallb/metallb project
```
oribon: change this TODO to be about extending MetalLB's FRR.Route struct, and open an issue to track this so we won't forget 😅

e2etests/tests/advertisement.go
```diff
@@ -468,6 +469,78 @@ var _ = ginkgo.Describe("Advertisement", func() {
 				},
 				splitCfg: splitByLocalPref,
 			}),
+			ginkgo.Entry("IPV4 - Advertise with asPathPrepend, AllowedMode all", params{
+				ipFamily: ipfamily.IPv4,
+				vrf:      "",
+				myAsn:    infra.FRRK8sASN,
+				prefixes: []string{"192.168.0.0/24", "192.168.1.0/24"},
```
oribon: can we add another prefix and validate it does not get a prepend?

e2etests/tests/advertisement.go
```diff
@@ -468,6 +469,78 @@ var _ = ginkgo.Describe("Advertisement", func() {
...
+									Prefixes:      []string{"192.168.1.0/24"},
+								},
+							}
+						}
+					}
+				},
+				validate: func(ppV4 []config.Peer, ppV6 []config.Peer, nodes []v1.Node) {
+					for _, p := range ppV4 {
+						// Only validate eBGP neighbors
+						if strings.Contains(p.FRR.Name, "ebgp") {
```
oribon, line 498: not sure I understand this assumption (or where it comes from), the api allows setting it for both ibgp and ebgp so why do we cover only ebgp?

api/v1beta1/frrconfiguration_types.go
```diff
@@ -355,6 +361,31 @@ type CommunityPrefixes struct {
...
+	// (e.g., creating active/standby routing).
+	//
+	// For example, if your ASN is 65000 and you set this to 3, the
+	// controller configures FRR to 'set as-path prepend 65000 65000 65000'.
+	// If a specific LocalASN is configured for the neighbor, that LocalASN
+	// is used instead to ensure BGP masquerading remains intact.
+	//
+	// We intentionally restrict this to repeating the local ASN. This approach
+	// provides the benefits of a longer AS-path while protecting users from
+	// accidentally triggering upstream loop-prevention drops.
```
oribon: this is too long, can you narrow it down please (like localpref/community)?

oribon (commented):
pretty much lgtm, just a couple of nits
can you also organize (squash) the commits so that it is ready for merge?

e2etests/tests/validate.go
psasnal: Ok, I initially didn't understand what you meant :) comment fixed!

e2etests/tests/advertisement.go
psasnal: done

e2etests/tests/advertisement.go
psasnal: The API and controller allow setting `PrefixesWithAsPathPrepend` on both iBGP and eBGP neighbors. By design, frr-k8s lets the controller faithfully translate user intent without blocking configurations that might be ineffective at the protocol level—matching how FRR itself operates and how we handle other attributes (like `LocalPref`).

In standard BGP behavior, AS-Path prepending is an eBGP-specific traffic engineering tool. When frr-k8s prepends its local ASN to an iBGP peer, the receiving peer sees its own ASN in the path and drops the route due to standard BGP loop prevention.

I looked into enabling the `allowas-in` setting on the `ibgp-multi-hop` test container to cover iBGP in e2e, but FRR strictly restricts `allowas-in` to eBGP sessions. Because receiving iBGP peers drop these prepended routes by design in FRR, the e2e test validates AS-Path prepend strictly for eBGP neighbors.

e2etests/tests/validate.go
psasnal: Done! Opened issue metallb/metallb#3120 to track extending `FRR.Route` and updated the TODO comment in `validate.go` to reference it.

e2etests/tests/validate.go
```diff
@@ -175,6 +176,69 @@ func ValidateNeighborLocalPrefForPrefix(neigh frrcontainer.FRR, prefix string, e
 	}, 5*time.Second, time.Second).ShouldNot(HaveOccurred())
 }
 
+func ValidateNeighborAsPathPrependForPrefix(neigh frrcontainer.FRR, prefix string, expectedAsPathPrepend uint8, ipfam ipfamily.Family) {
+	if !strings.Contains(neigh.Name, "ebgp") {
```
fedepaol, line 180: I would remove this from here and be intentional on the calling site. This function should do what it advertises, which is validate that the neighbor has aspath prepend

e2etests/tests/validate.go
```diff
@@ -175,6 +176,69 @@ func ValidateNeighborLocalPrefForPrefix(neigh frrcontainer.FRR, prefix string, e
...
+
+		if asPathPrepend != expectedAsPathPrepend {
+			return fmt.Errorf("asPathPrepend %d for prefix %s on neighbor %s does not equal %d", asPathPrepend, prefix, neigh.Name, expectedAsPathPrepend)
+		}
+
+		return nil
+	}, 5*time.Second, time.Second).ShouldNot(HaveOccurred())
+}
+
+// TODO(metallb/metallb#3120): Extend MetalLB's FRR.Route struct to natively parse and store AS-Path prepend data.
```
fedepaol, line 199: this todo shouldn't be here

e2etests/tests/validate.go
```diff
@@ -175,6 +176,69 @@ func ValidateNeighborLocalPrefForPrefix(neigh frrcontainer.FRR, prefix string, e
...
+func AsPathPrependForPrefix(neigh frrcontainer.FRR, prefix string, ipfam ipfamily.Family) (uint8, error) {
+	// 1. Run the command inside the external FRR container
```
fedepaol, line 201: no need for order number. Also, we do this kind of operations in https://github.com/metallb/frr-k8s/tree/main/e2etests/pkg/frr , I would be consistent with that

e2etests/go.mod
```diff
@@ -64,16 +64,16 @@ require (
 	go.universe.tf/metallb v0.15.2 // indirect
 	go.yaml.in/yaml/v2 v2.4.2 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
-	golang.org/x/net v0.53.0 // indirect
-	golang.org/x/oauth2 v0.30.0 // indirect
-	golang.org/x/sync v0.20.0 // indirect
-	golang.org/x/sys v0.43.0 // indirect
-	golang.org/x/term v0.42.0 // indirect
-	golang.org/x/text v0.36.0 // indirect
+	golang.org/x/net v0.56.0 // indirect
```
fedepaol, line 67: are these bumps required for this PR

internal/controller/merge.go
```diff
@@ -141,8 +141,26 @@ func mergeAllowedOut(r, toMerge frr.AllowedOut) (frr.AllowedOut, error) {
 		}
 	}
 
+	asPathPrependForPrefix := map[string]string{}
+	for _, p := range r.AsPathPrependPrefixesModifiers {
+		// Convert to a single string for comparison (format '<PrependASN> x <PrependCount>')
+		prependStr := fmt.Sprintf("%s x %d", p.PrependASN, p.PrependCount)
```
fedepaol, line 147: use a function to get the key here

internal/controller/merge.go
```diff
@@ -141,8 +141,26 @@ func mergeAllowedOut(r, toMerge frr.AllowedOut) (frr.AllowedOut, error) {
...
+		for _, prefix := range p.Prefixes.UnsortedList() {
+			asPathPrependForPrefix[prefix] = prependStr
+		}
+	}
+	for _, p := range toMerge.AsPathPrependPrefixesModifiers {
+		prependStr := fmt.Sprintf("%s x %d", p.PrependASN, p.PrependCount)
```
fedepaol, line 153: and here

internal/controller/merge.go
```diff
@@ -141,8 +141,26 @@ func mergeAllowedOut(r, toMerge frr.AllowedOut) (frr.AllowedOut, error) {
 		}
 	}
 
+	asPathPrependForPrefix := map[string]string{}
```
fedepaol, line 144: can we add a comment with the merge logic here? IIUC the lists must be identical, but this can (and should) be discussed

internal/controller/merge_test.go
```diff
@@ -2008,19 +2226,24 @@ func communityPrefixListSorter(a, b frr.CommunityPrefixList) bool {
 		panic("empty name")
 	}
 
-	if communityPrefixListKey(a.Community, a.IPFamily) < communityPrefixListKey(b.Community, b.IPFamily) {
-		return false
-	}
-	return true
+	return communityPrefixListKey(a.Community, a.IPFamily) <
```
fedepaol, line 2229: not related to this change

internal/controller/merge_test.go
```diff
@@ -2008,19 +2226,24 @@ func communityPrefixListSorter(a, b frr.CommunityPrefixList) bool {
...
 }
 
 func localPrefPrefixListSorter(a, b frr.LocalPrefPrefixList) bool {
 	if a.Name == "" || b.Name == "" {
 		panic("empty name")
 	}
 
-	if localPrefPrefixListKey(a.LocalPref, a.IPFamily) < localPrefPrefixListKey(b.LocalPref, b.IPFamily) {
-		return false
+	return localPrefPrefixListKey(a.LocalPref, a.IPFamily) <
```
fedepaol, line 2238: same

fedepaol (changes requested):

api/v1beta1/frrconfiguration_types.go
```diff
@@ -355,6 +361,19 @@ type CommunityPrefixes struct {
 	Community string `json:"community,omitempty"`
 }
 
+// AsPathPrependPrefixes is a list of prefixes associated with an AS path prepending.
+type AsPathPrependPrefixes struct {
+	// Prefixes is the list of prefixes associated with the AS path prepending.
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:Format="cidr"
+	Prefixes []string `json:"prefixes,omitempty"`
```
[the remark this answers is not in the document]
psasnal: Fixed

internal/controller/merge_test.go
psasnal: I originally touched it because the existing logic violates Go's strict weak ordering, but I agree it's out of scope for this feature. Reverted, this minor fix can be merged as a part of a separate PR in the future.

internal/controller/merge_test.go
psasnal: see above

e2etests/tests/validate.go
psasnal: Ok!

e2etests/tests/validate.go
psasnal: I originally modeled this after LocalPrefForPrefix (which relies on the metallb/metallb testing package) and left the TODO to implement it later based on an earlier discussion with @oribon. I've moved the vtysh parsing logic to https://github.com/metallb/frr-k8s/tree/main/e2etests/pkg/frr/frr.go as requested. Unnecessary numbers also removed from comments.

internal/controller/merge.go
psasnal: Ah, good catch. I've updated this to reuse asPathPrependPrefixListKey to generate the comparison string, keeping it consistent with the helpers used elsewhere.

e2etests/go.mod
psasnal: No, they aren't required here. The e2etests bumps were just an accidental side-effect of running go mod tidy locally. I've reverted e2etests/go.mod, e2etests/go.sum, and e2etests/go.work.sum to latest main branch.

internal/controller/merge.go
psasnal: Added the comment! To clarify the logic: the lists (`r` and `toMerge`) themselves don't need to be identical, but if two modifiers target the exact same prefix, their prepend values (ASN and count) must be identical. I mirrored this directly from the LocalPref conflict resolution logic right above it. If two different configs try to dictate different prepend counts for the same IP, it's an ambiguous routing conflict, so erroring out seems like the safest approach.

api/v1beta1/frrconfiguration_types.go
```diff
@@ -355,6 +361,19 @@ type CommunityPrefixes struct {
...
+
+	// AsPathPrepend is the number of times to prepend the local ASN to the AS path for the prefixes.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=10
+	AsPathPrepend uint8 `json:"asPathPrepend,omitempty"`
```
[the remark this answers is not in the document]
psasnal: Fixed

e2etests/tests/validate.go
fedepaol: this was resolved but the comment is still there

e2etests/tests/validate.go
fedepaol: not sure I understand. The unmarshaling is still here, I'd move it under the frr package together with the others.

e2etests/go.mod
fedepaol: I still see the changes here?

go.mod
```diff
@@ -23,7 +23,7 @@ require (
 )
 
 require (
-	cel.dev/expr v0.19.1 // indirect
+	cel.dev/expr v0.25.1 // indirect
```
fedepaol: same here for deps

e2etests/tests/validate.go
psasnal: @fedepaol I accidentally marked a few comments as resolved while working through them locally. I'm batching up the rest of the feedback now and will push all the fixes up together shortly!

psasnal: TODO removed

e2etests/tests/validate.go
psasnal: committed pending changes

e2etests/tests/validate.go
```diff
@@ -175,6 +176,69 @@ func ValidateNeighborLocalPrefForPrefix(neigh frrcontainer.FRR, prefix string, e
...
+		return 0, nil // No ASNs in the path
+	}
+
+	// 5. Count the ASNs.
+	// If the base ASN is added once natively, and we prepend 3 times, there are 4 ASNs total.
+	// So prependCount = Total ASNs - 1
+	asnList := strings.Fields(asPathStr)
+	prependCount := len(asnList) - 1
+
+	return uint8(prependCount), nil
```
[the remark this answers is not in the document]
psasnal: Fixed

e2etests/go.mod
psasnal: Committed the change now

e2etests/tests/advertisement.go
```diff
@@ -468,6 +469,80 @@ var _ = ginkgo.Describe("Advertisement", func() {
...
 			}),
+			ginkgo.Entry("IPV4 - Advertise with asPathPrepend, AllowedMode all", params{
+				ipFamily: ipfamily.IPv4,
+				vrf:      "",
+				myAsn:    infra.FRRK8sASN,
+				prefixes: []string{"192.168.0.0/24", "192.168.1.0/24", "192.168.2.0/24"},
+				modifyPeers: func(ppV4 []config.Peer, ppV6 []config.Peer) {
+					for i := range ppV4 {
+						// Only apply AsPathPrepend to eBGP neighbors to avoid loop-prevention drops on iBGP
+						if strings.Contains(ppV4[i].FRR.Name, "ebgp") {
```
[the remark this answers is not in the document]
psasnal: I've added `iseBGPPeer` helper function. I've also added a counter and an assertion to guarantee the test fails if the topology changes and no eBGP peers are found (preventing silent passes).

internal/controller/api_to_config.go
```diff
@@ -430,6 +439,53 @@ func prefixesWithCommunityToFRR(toAdd map[string]frr.CommunityPrefixList, neighb
...
+				IPFamily: frrFamily,
+				Prefixes: sets.New[string](),
+			},
+			PrependASN:   asnToPrepend,
+			PrependCount: prependCount,
+		}
+
+		ipfamilyPrefixes := ipfamily.FilterPrefixes(prefixes.Prefixes, ipFamily)
+		if len(ipfamilyPrefixes) == 0 {
+			continue
```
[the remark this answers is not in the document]
psasnal: check and unit tests added

internal/controller/api_to_config.go
```diff
@@ -430,6 +439,53 @@ func prefixesWithCommunityToFRR(toAdd map[string]frr.CommunityPrefixList, neighb
...
+
+		asnToPrepend := fmt.Sprintf("%d", routerASN)
+		if neighbor.LocalASN > 0 {
+			asnToPrepend = fmt.Sprintf("%d", neighbor.LocalASN)
+		}
+
+		key := asPathPrependPrefixListKey(asnToPrepend, prependCount, frrFamily)
+		if _, ok := toAdd[key]; ok {
+			return nil, fmt.Errorf("AS path prepending %d is already defined", prefixes.AsPathPrepend)
+		}
```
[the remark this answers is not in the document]
psasnal: I added this logic to `validateOutgoingPrefixes` function, which now validates the address families and rejects conflicting prepend counts before the configuration ever reaches this generation step. Also, unit test for this scenario was added.

go.mod
psasnal: Some of the packages were bumped because of vuln check, but maybe this is not needed anymore.  I've completely reverted the root go.mod and go.sum to keep the PR strictly focused on the feature. If the CI security scanner flags anything specific on this run, I'll only bump the exact package it requires.

internal/controller/merge_test.go
```diff
@@ -1963,6 +2166,9 @@ func TestMergeEVPNConfigs(t *testing.T) {
 			if test.err == nil && err != nil {
 				t.Fatalf("expected no error, got %v", err)
 			}
+			if test.err != nil && err != nil && !strings.Contains(err.Error(), test.err.Error()) {
+				t.Fatalf("expected error containing %q, got %q", test.err.Error(), err.Error())
+			}
```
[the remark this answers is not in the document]
psasnal: Removed unnecessary return and added error message comparison (plus fixed tests because of that).

psasnal (commented):
@fedepaol Now all the changes are committed, sorry for confusion.

PR #462, opened 2026-06-30
Add AllowAsIn field to Neighbor FRRConfiguration

jcaamano:
Adds support for the BGP `allowas-in` neighbor directive via a new `AllowAsIn` field on the `Neighbor` configuration. This is useful in hub-and-spoke or route-leaking topologies where the same AS number may appear multiple times in the path.

/kind feature

  ### Values

  - `""` (empty, default) or `"none"`: routes with the local AS in the path are rejected.
  - `"origin"`: routes are accepted only if the local AS appears as the origin (last AS in the path).
  - `"1"-"10"`: routes are accepted with the local AS appearing as much as that many times in the path.

  The directive is applied to all enabled address families (ipv4/ipv6 unicast and l2vpn evpn).

  ### Merge behavior

  When multiple configurations target the same neighbor, `"none"` explicitly prevents any other configuration from enabling allowas-in. Any other combination of values resolves to the least restrictive.

  ## Changes

  - API: new `AllowAsInMode` enum type and `AllowAsIn` field on `Neighbor`
  - Templates: `allowas-in` directive rendered in `neighboripfamily.tmpl` and `evpn.tmpl`
  - Merge: `mergeAllowAsIn` implements the merge strategy
  - Tests: golden file tests (ipv4, ipv6 dual-stack, EVPN), merge tests, CRD validation tests
  - Generated: CRD manifests, Helm chart CRD, API docs

### DIFF
--- a/api/v1beta1/frrconfiguration_types.go
+++ b/api/v1beta1/frrconfiguration_types.go
@@ -229,6 +229,23 @@ type Neighbor struct {
 	// +kubebuilder:validation:Format=int64
 	LocalASN uint32 `json:"localASN,omitempty"`
 
+	// AllowAsIn controls whether routes with the local AS number in the AS path
+	// are accepted from this neighbor for the enabled address families.
+	// This is useful in hub-and-spoke or route-leaking topologies where the
+	// same AS number may appear multiple times in the path.
+	// Possible values:
+	// - "" (empty, default): routes with the local AS in the path are rejected.
+	//   Does not reject other AllowAsIn values targeting the same neighbor.
+	// - "none": routes with the local AS in the path are rejected.
+	//   Rejects other AllowAsIn values targeting the same neighbor.
+	// - "origin": routes are accepted only if the local AS appears as the origin (last AS in the path).
+	// - "1"-"10": routes are accepted with up to this many occurrences of the local AS in the path.
+	// When multiple configurations target the same neighbor, any combination of
+	// values resolves to the least restrictive, except "none" which rejects
+	// all other values.
+	// +optional
+	AllowAsIn AllowAsInMode `json:"allowAsIn,omitempty"`
+
 	// AddressFamilies specifies which address families to activate this neighbor for.
 	// Supported values: "unicast" (IPv4/IPv6 unicast based on neighbor IP), "evpn" (L2VPN EVPN).
 	// +optional
@@ -466,6 +483,15 @@ type AddressFamily string
 const (
 	AddressFamilyUnicast AddressFamily = "unicast"
 	AddressFamilyEVPN    AddressFamily = "evpn"
+)
+
+// AllowAsInMode specifies whether routes with the local AS in the path are accepted from a neighbor.
+// +kubebuilder:validation:Enum="";none;origin;"1";"2";"3";"4";"5";"6";"7";"8";"9";"10"
+type AllowAsInMode string
+
+const (
+	AllowAsInNone   AllowAsInMode = "none"
+	AllowAsInOrigin AllowAsInMode = "origin"
 )
 
 // AdvertisePrefixType specifies a prefix type to advertise as EVPN type-5 routes.
--- a/e2etests/tests/evpn.go
+++ b/e2etests/tests/evpn.go
@@ -60,333 +60,210 @@ var evpnL3ConfigTemplate = infra.EVPNConfig{
 	Vxlan:           evpnVxlan,
 }
 
-var _ = ginkgo.DescribeTableSubtree("EVPN",
-	func(peerFamily ipfamily.Family, advertiseFamily ipfamily.Family) {
-		var (
-			cs               clientset.Interface
-			nodes            []corev1.Node
-			evpnCfg          infra.EVPNConfig
-			evpnInfra        *infra.EVPN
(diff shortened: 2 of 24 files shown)


jcaamano:
@fedepaol I realized the project follows a lower case enum value policy which is different from kubernetes camel case style. One of the Enums introduced in the EVPN PR used camel case. Let me know if we want to address that before bumping

api/v1beta1/frrconfiguration_types.go
```diff
@@ -229,6 +229,20 @@ type Neighbor struct {
...
+	// same AS number may appear multiple times in the path.
+	// Possible values:
+	// - "" (empty, default) or "none": routes with the local AS in the path are rejected.
+	// - "origin": routes are accepted only if the local AS appears as the origin (last AS in the path).
+	// - "any": routes are accepted regardless of how many times the local AS appears in the path.
+	// When multiple configurations target the same neighbor, "none" explicitly prevents
+	// any other configuration from enabling allowas-in. Any other combination of values
+	// resolves to the least restrictive.
+	// +optional
+	AllowAsIn AllowAsInMode `json:"allowAsIn,omitempty"`
```
fedepaol, line 247: I'd also (optionally) add the number of instances as per https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighbor-A.B.C.D-X-X-X-X-WORD-allowas-in-1-10-origin
In that case it might make sense to have a struct out of this

fedepaol (commented):
nice and straightforward!
Is there a way to e2e this too?

api/v1beta1/frrconfiguration_types.go
jcaamano: I was thinking the potential future expansion would be to just open the values.

BTW I thought not specifying a number would equate to any number of times but it seams it doesn't:
```
The parameter (1-10) configures the amount of accepted occurrences of the system AS number in AS path for matching routes (default: 3).
```
There no option rally to express `any`
So I think this makes sense:
```
// +kubebuilder:validation:Enum="";none;origin;1;2;3;4;5;6;7;8;9;10
```
If the value expands in the future to be unbounded we can always change the CEL validation to something else.

I don't think a struct is needed because it is essentially a single value?

jcaamano: > BTW I thought not specifying a number would equate to any number of times but it seams it doesn't:
> 
> ```
> The parameter (1-10) configures the amount of accepted occurrences of the system AS number in AS path for matching routes (default: 3).
> ```

Wait, that's specific to the route map format. Hmmm that's confusing...

jcaamano: Yeah, I think the default is 3

api/v1beta1/frrconfiguration_types.go
```diff
@@ -236,7 +236,7 @@ type Neighbor struct {
 	// Possible values:
 	// - "" (empty, default) or "none": routes with the local AS in the path are rejected.
```
oribon: can we have these separated? it might read as empty or none being the same, but none has the additional semantic of denying merges with anything else (that is not empty)

e2etests/tests/evpn.go
```diff
@@ -61,7 +60,7 @@ var evpnL3ConfigTemplate = infra.EVPNConfig{
 }
 
 var _ = ginkgo.DescribeTableSubtree("EVPN",
-	func(peerFamily ipfamily.Family, advertiseFamily ipfamily.Family) {
+	func(containerName string, peerFamily ipfamily.Family, advertiseFamily ipfamily.Family, allowAsIn frrk8sv1beta1.AllowAsInMode) {
```
oribon: do we have to modify the existing test instead of having the new tests separate (like in the allowas_in file)?

jcaamano: The reason why unicast tests are separated is because the pre-configuration of the test is different when compared to other tests because most of the payload is configuring the external container to prepend ASN, so it wouldn't share much of that code preparation and did not fit any existing test.

However for EVPN preparing the infra is a lot more of the test payload which also requires configuring the external containers in a special way so a lot of that can be shared across tests.

oribon: not sure I completely follow (didn't catch up with the evpn changes yet), but to me it seems that we're forcing this change on existing tests (where the first 3 table cases stay as-is) to save some lines, which isn't really maintainable/readable in the long run (e.g what happens when the next feature needs evpn testing).
what's the cost of splitting this into its dedicated Table/It versus "forcefully adapting" the existing test to support this?
also, in the existing implementation we leave the negative test outside (I assume because it doesn't look good), which is another weak point of the latter.

jcaamano: Unicast test was not going to share much of the code with existing tests, so its a new test.
EVPN test shares 99% of the code with the existing test, so uses existing test.

The negative test is basically there as an FRR expectation test, to check that external container prepend is behaving as expected, is not really testing anything important over an existing positive test.

Anyway, will try to comply.

jcaamano: Reverted test to original form and added commit with new test, to be squashed.

oribon (approved):
lgtm thanks!

